PR #6, opened 2018-09-10
feat(responsive-background): Make the website's background image responsive

WNemencha:
Closes #4.


src/styles/scss/index.scss
```diff
@@ -1,4 +1,8 @@
-@import 'base/variables';
-@import 'base/elements';
+@import "base/variables";
```
crysfel: I think we need to decide whether we go with double/single quotes, for consistency.

codinronan: I (and I think most - airbnb rule) use single for JS, double for JSX, as this retains consistency with HTML.

src/App.js
```diff
@@ -1,27 +1,26 @@
...
-            </a>
-          </button>
-        </header>
+        <section className="App__hero">
+          <header className="App__header">
+            <h1 className="App__title">Coding Coach</h1>
+            <h4 className="App__tagline">
+              Connecting developers with mentors worldwide.
+            </h4>
+            <button className="button button__white--fill">
```
AngeloGulina: not a big deal, but probably should have a different name for this `button__white--fill`, if you're following BEM (as it seems). Both `white` and `fill` seem modifiers. What do you think?

src/App.js
```diff
@@ -1,27 +1,26 @@
...
-            <a 
-              href="https://docs.google.com/forms/d/1TraoygcJay39cLiBQM-ROlToy7xIgcdx_VvCQuE1lJw/edit#responses"
-              target="blank"
-              className="button__link"
-            >
-              Get Started
-            </a>
-          </button>
-        </header>
+        <section className="App__hero">
```
AngeloGulina: What do you think if we switch to lowercase early on?

src/styles/scss/index.scss
AngeloGulina: How do you gals/guys approach imports?
I always make a case to import things such `variables` on every file, where needed instead the "main" file.
What do you think?

src/styles/css/index.css
```diff
@@ -38,15 +38,14 @@ body, html {
 .App {
```
AngeloGulina: Generated file should go probably in a `dist` dir and being ignored. What do you think?

src/App.js
reeversedev: Not really sure but why are we using class names with double underscores?

src/App.js
emmawedekind: For components? They should be uppercase.

src/App.js
emmawedekind: I agree white & fill are both modifiers.

emmawedekind:
You have to remove `padding: 24px` on `<div className="App"`.  Otherwise you get x and y scrollbars.

SkyzohKey:
@emmawedekind I thought I had removed it lol, will do tomorrow, I'm sick for the moment, not enough motivation to open my laptop sorry 😖...

src/App.js
AngeloGulina: You mean writing CSS classes upper case, because it's a "component"? (Just asking, never heard of this approach).

WNemencha: Yeah. When you defines styles for a component, capitalize the definition so it is consistent with how its written in JSX :p

src/styles/css/index.css
WNemencha: It already is ignored. I just removed the two files as they are not needed anymore.

emmawedekind:
Awesome, could you just fix this incompatibility with es-lint please? :) 
image

WNemencha:
@emmawedekind How to reproduce the error ? I did an `yarn && yarn start` and it works (except for the `Configuration for rule "react/jsx-no-bind" is invalid` but this is out of scope for this PR)

emmawedekind:
I see issues in the terminal still, I've messaged you on the help channel. If we figure this out, I'll merge it.

PR #9, opened 2018-09-10
Feature/i18n support

Halkreen:
### Description

Added i18n support for CodingCoach.

### New dependencies

"i18next": "^11.7.0",
"i18next-browser-languagedetector": "^2.2.3",
"react-i18next": "^7.12.0"

### Implementation

- Main app container is now **inside the I18nextProvider container**
- The translate method has been imported into the landing page as `t()`
- The translation keys are located inside `src/i18n/i18n.js`
- Default language is the one detected through browser or locale.

### Improvements

- It might be wiser for clarity sake to create individual files for each language (having a i18n.js and en.json, fr.json, de.json containing only the "raw" translations).
- A dropdown at the top-right corner to switch language.

(these 2 points can be new issues)


src/index.js
```diff
@@ -1,10 +1,15 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
+import { I18nextProvider } from "react-i18next";
```
crysfel: For code consistency, let's stick with single quotes for JavaScript code and double quotes for JSX.

src/index.js
```diff
@@ -1,10 +1,15 @@
...
 
+import i18n from "./i18n/i18n";
```
crysfel: I'd prefer to have this inside the `App` component. Later on we are going to refactor that component to include the redux store provider, the router, etc... I'd like to have everything in a single place. What do you guys think?

src/i18n/i18n.js
```diff
@@ -0,0 +1,41 @@
+import i18n from "i18next";
+import LanguageDetector from "i18next-browser-languagedetector";
+
+i18n.use(LanguageDetector).init({
+  // we init with resources
+  resources: {
+    en: {
```
crysfel: This approach is ok for now, maybe add a @TODO to split the translations text in different files ;)

src/index.js
codinronan: Having used i18next extensively, in fact we'll want both this and the I18nextProvider to be in the `App` component. They'll both need information that the store + app have access to.

src/index.js
anteloe: Do we want to use prettier for this and on pre-push, automatically apply it to all files?

ejbostian: Sure! @anteloe go for it.

emmawedekind:
Looks great!

emmawedekind:
Let's add the css.map files to the .gitignore and if you could rebase with develop so there are no merge conflicts, I'll go ahead and merge!

Halkreen:
I merged with development, but it was on another branch of my fork, can you close this PR and merge the other one instead ? Only difference is the development rebase 

https://github.com/Coding-Coach/coding-coach/pull/24

crysfel:
Closing due to duplication

PR #13, opened 2018-09-11
User can click button instead of a tag

hakankaraduman:
changed click action from a tag to button, so that user does not have to click text. Earlier the click on button wasn't working but only the a tag was working.


src/App.js
```diff
@@ -1,6 +1,11 @@
 import React, { Component } from 'react';
 
 class App extends Component {
+  onGetStarted = () => {
+    const url = 'https://docs.google.com/forms/d/1TraoygcJay39cLiBQM-ROlToy7xIgcdx_VvCQuE1lJw/edit#responses';
```
AngeloGulina: First of all, great catch!
Second, unless there is a specific reason or behavior you want to obtain I would not use this approach.
There is far way easier way to obtain this via simple CSS, working with `padding` and such targeting the `a` as a block element.

Beside that, CSS comes with the advantage of setting up a default style for this kind of action – so, in future, all buttons would look the same and have ideally one single point from where style and behavior is defined.

nero2009:
Not sure if this is relevant here but this stackflow answer said we should use a tags for navigation and button tags for submit actions. 
https://stackoverflow.com/questions/21715162/when-should-i-use-a-button-button-or-a-link-a-in-html

src/App.js
```diff
@@ -17,11 +12,9 @@ class App extends Component {
         <header className="App__header">
           <h1 className="App__title">Coding Coach</h1>
           <h4 className="App__tagline">Connecting developers with mentors worldwide.</h4>
-          <button className="button button__white--fill" onClick={this.onGetStarted}>
-            <a className="button__link">
-              Get Started
-            </a>
-          </button>
+          <a className="button button__white--fill button__link">
```
palashmon: Hi, thanks for your contribution. But, it seems like this `<a>` element is missing the `href` attribute now which previously pointed to the google docs. Could you please verify that once?

Regards!

hakankaraduman: I forgot to add back, fixed.

ejbostian:
Let me know when this is ready for review!

palashmon:
LGTM :+1:

yarn-error.log
```diff
@@ -1,5 +1,6 @@
 Arguments: 
-  C:\Program Files\nodejs\node.exe C:\Program Files (x86)\Yarn\bin\yarn.js sass
```
AngeloGulina: We can ignore this one.

src/App.js
AngeloGulina: It's totally fine to commit on the PR again, sure!
Great job and also great reaction time 😄

yarn-error.log
hakankaraduman: Should I add this as a commit on this pull request?

hakankaraduman: I added to the .gitignore those files, but it didn't work I think

AngeloGulina: If I'm not mistaken, you need to run `git rm --cached [[you file]]`.

palashmon: Yup, its a known issue. Here is a solution.

https://stackoverflow.com/a/1139797/1823841

hakankaraduman: Thanks so much, done and committed

Odonno:
Looks good to me.

PR #18, opened 2018-09-11
updated react-scripts to v2 to support css modules

nero2009:


package.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "react": "^16.5.0",
     "react-dom": "^16.5.0",
-    "react-scripts": "1.1.5"
+    "react-scripts": "^2.0.0-next.a671462c"
```
jagzviruz: I am not a huge fan of using `next` versions as they tend to make upgrading painful. In recent times, I had real issues when I used material-ui@next for some of my projects.

crysfel:
After merging https://github.com/Coding-Coach/coding-coach/pull/16/files is not needed to upgrade to V2 to add support for css modules.

Let's discuss the benefits of upgrading to the latest.

emmawedekind:
I'm going to close this PR due to the conversation above. Let me know if it needs to be re-opened.

PR #27, opened 2018-09-12
FIx: Added additional formatting rules for editors

jagzviruz:
Added `.editorconfig` and updated `.prettierrc` rules


palashmon:
Looks good to me :+1:

PR #25, opened 2018-09-12
Fix: `test` script

jagzviruz:
Changed `test` script to use original CRA command with `react-scripts`

PR #28, opened 2018-09-12
Added Travis integration config.

jagzviruz:
Travis config has been added and tested with forked repository.
The flag can be seen at `https://github.com/jagzviruz/coding-coach/tree/feature/travis` however it says 'unknown' because the url is pointed to 'coading-coach'.

### Next steps
@emmawedekind has to integrate the repo with [Travis](https://travis-ci.org) to run build on the development branch


DennisKo:
add `- npm run test` as well

.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+  - "8"
+cache:
+  directories:
+    - node_modules
+before_install:
+  - npm install -g greenkeeper-lockfile@1
+install: npm install
+before_script:
+  - greenkeeper-lockfile-update
+script:
```
Asjas: Perhaps we should add the `npm run build` command here as well?

So that we can make sure that both the `linting` works and that the website can be built without any build errors. So the linting command would run first and then the website will build second.

crysfel: Yeap, totally agree. Let's run the linter, tests and make sure is building correctly.

Asjas: Yeah I forgot about adding the `tests` command as well. :smile: 

I agree with that order. `linting`, `tests` and then `build`.

jagzviruz: > I'm ok with TravisCI, @jagzviruz can you add to our tech stack document why Travis over CircleCI?
> 
> https://docs.google.com/document/d/1DbdRZBsyjbeBl9vddUyXCUWsgwdgyyUY2NIQitf3_Z0/edit#heading=h.e0bl2e8hwksl
> 
> Thanks!

Will update it later in my evening.

guillermaster:
> Have we decided to use Travis CI over Circle CI?

According to one of the tech stack documents, I thought we were going to use CircleCI.

crysfel:
There are some conflicts in the `package.json` file, I will merge after resolving those conflicts!

jagzviruz:
The conflict is because of a PR which has reverted #25 ..

.travis.yml
nitish24p: > Perhaps we should add the `npm run build` command here as well?
> 
> So that we can make sure that both the `linting` works and that the website can be built without any build errors. So the linting command would run first and then the website will build second.

If we're doing npm build just for the sake of linting, then the eslint loader of webpack will anyway be run the lint before the build.. so will be redundant acc to me

package.json
```diff
@@ -16,6 +16,7 @@
     "start": "node scripts/custom-config start",
     "build": "node scripts/custom-config build",
     "deploy": "gh-pages -d build",
+    "lint": "eslint src/",
```
nitish24p: i think you should add a regex for only .js files cause this will even lint the scss files and may throw errors

.travis.yml
palashmon: Yup, also `lint-staged` is running before every commit. So, is it like 3 times linting will happen?

arthurbuhl:
I think this PR is good to be merged. Please someone with write access approve it. Thanks.

emmawedekind:
Approved. I'll merge now.

PR #31, opened 2018-09-12
Proposal for react-testing-library

DennisKo:
React Testing Library (RTL) is a new testing library quite similar to enzyme. It encourages better testing practices by getting our tests closer to using our components the way a user will. It has gained a lot of traction the last months and is [mentioned in the official React docs](https://reactjs.org/docs/test-utils.html#overview). Its API is similar enough to enzyme for people new to RTL to get used to it quickly.

The authors article about it:
https://www.google.com/url?q=https://blog.kentcdodds.com/introducing-the-react-testing-library-e3a274307e65&sa=D&ust=1536735772169000&usg=AFQjCNGsoIg3fGsPgEaTX1TZ3R28tU1INQ

The library itself:
https://github.com/kentcdodds/react-testing-library

Discussion: Pros/Cons of RTL & Enzyme?

### DIFF
--- a/src/setupTests.js
+++ b/src/setupTests.js
@@ -1,0 +1,6 @@
+// react-testing-library renders your components to document.body,
+// this will ensure they're removed after each test.
+import 'react-testing-library/cleanup-after-each';
+
+// this adds jest-dom's custom assertions
+import 'jest-dom/extend-expect';
--- a/package-lock.json
+++ b/package-lock.json
@@ -1493,7 +1493,7 @@
     },
     "browserify-aes": {
       "version": "1.2.0",
-      "resolved": "http://registry.npmjs.org/browserify-aes/-/browserify-aes-1.2.0.tgz",
+      "resolved": "https://registry.npmjs.org/browserify-aes/-/browserify-aes-1.2.0.tgz",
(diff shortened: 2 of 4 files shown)


dbanbahji:
I really like it, i have used it. Kent (the creator) has a great youtube channel with a lot of workshops at paypal about it teaching it to many teams.

src/setupTests.js
```diff
@@ -0,0 +1,6 @@
+// react-testing-library renders your components to document.body,
```
crysfel: Let's create a new folder to keep all the `test` configurations, mocks, etc... in a single place. Please create a `tests/setup` folder at the root level and place this file in there.

```
+ root
  + scripts
  + src
  + tests
    + setup
    + mocks
    + reports
```

DennisKo:
```
Out of the box, Create React App only supports overriding these Jest options:

  • collectCoverageFrom
  • coverageReporters
  • coverageThreshold
  • snapshotSerializers.

These options in your package.json Jest configuration are not currently supported by Create React App:

  • setupTestFrameworkScriptFile

If you wish to override other Jest options, you need to eject from the default setup. You can do so by running npm run eject but remember that this is a one-way operation. You may also file an issue with Create React App to discuss supporting more options out of the box.
```

Urghs.. :(

We really need more freedom in configuring the app!

DennisKo:
updated and ready to merge

PR #64, opened 2018-09-16
Internal: Setup Storybook config

arecvlohe:
## Description

This PR sets up the config for storybook. This will allow a developer to create a file `ComponentName.stories.js` anywhere in the project and it will create a story for it. I was not able to pull in the styles for a story, yet, but I wanted to wait since we are going to be using css modules which should make importing styles much easier.

To test out this PR, checkout the branch locally, install deps, and run `yarn storybook` to see the one story under `App`. 

## Screenshot

screen shot 2018-09-15 at 10 23 06 pm


arecvlohe:
I am having issues with sass. I will need to take some time to figure it out.

src/containers/App/App.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { func } from 'prop-types';
 import { translate } from 'react-i18next';
-import 'i18n/i18n';
+import '../../i18n/i18n';
```
crysfel: Is there a way to configure storybook to resolve the `src` folder? Adding those extra `../../` will be painful when adding more components later on

src/containers/App/assets/app.scss
```diff
@@ -1,8 +1,9 @@
-@import '~theme/base/variables';
+@import '../../../theme/base/variables';
```
crysfel: I'm assuming this is because storybook? Would it be possible to make storybook to resolve the `src` folder? 🤔

.storybook/webpack.config.js
```diff
@@ -0,0 +1,21 @@
...
+      test: /\.(png|jpg|gif)$/,
+      use: [
+        {
+          loader: 'file-loader',
+        },
+      ],
+    },
+  );
+
+  return storybookBaseConfig;
```
crysfel: We might be able to resolve the `src` folder here, right? Adding the following code seems to work 👌

```
storybookBaseConfig.resolve.modules.push('src');
```

arecvlohe: I will give this a try

src/containers/App/assets/app.scss
arecvlohe: It's also a matter of resolving the `~`. I will need to look into it some more

emmawedekind:
This is great!

arecvlohe:
@crysfel Fixed the issue with module resolving with your solution. Thanks! I think this PR is ready to be merged.

Issue #50, opened 2018-09-13
Google Analytics


emmawedekind:
**Acceptance Criteria:**
- Google Analytics are implemented in order to track user's interactions.

edward-hong:
I'm interested in implementing this feature

edward-hong:
Hi, I've imported `react-ga` and initialized it on render in `App.js`
```
 ReactGA.initialize(trackingID);
 ReactGA.pageview(window.location.pathname);
```
What should I do about the actual tracking ID?

nitish24p:
pick up the tracking id from an ENV variable, also you can abstract all the functionality to `analytics.js ` or some file like that

edward-hong:
Thanks @nitish24p I've wired up ENV variables with `dotenv` to `webpack` and have abstracted it to a separate file. Can I change App.js to a class component? It seems overkill to initialize `ReactGA` in the render function as it will be re-initialized every rerender. Wouldn't be more appropriate to initialize at `componentDidMount()`?

PR #70, opened 2018-09-21
Feature/#50 Google Analytics

edward-hong:
# Details

## Description

The aim of this PR is to add Google Analytics. The following is a list of steps I took to achieve this.

1. Then I added some code to `scripts/custom-config.js`. This is so Webpack can compile the `process.env` variables so we can use them on the front end.
2. Created a file called `googleAnalytics.js` in the `src` directory that exports two functions. One initializes the Google Analytics and the second track the current page.
3. Rewrote `App.js` into a class component. The reason I did this was that I thought the best way to run the two functions was during `componentDidMount` life cycle hook.

~~Added `.env.example` file with a list of variables~~
~~Made a small amendment to README.md about `.env.example`~~

refer to #81 regarding `process.env`

## Issue

[#50](https://github.com/Coding-Coach/coding-coach/issues/50) Google Analytics


scripts/custom-config.js
```diff
@@ -71,5 +73,18 @@ function rewireModule(modulePath, customizer) {
   // react-scripts imports the config as a `const` and we can't
   // modify that reference.
   let config = defaults.__get__('config');
+
+  // Have webpack compile the env keys so the can be used on the client
+  var env = dotenv.config().parsed;
```
linux-nerd: @edward-hong CRA has inbuilt configuration for .env file. If you have your keys renamed to REACT_APP_GA_TRACKING_ID in .env file, you can access the same via process.env.REACT_APP_GA_TRACKING_ID from js.

src/googleAnalytics.js
```diff
@@ -0,0 +1,4 @@
+import ReactGA from 'react-ga';
+
+export const initializeGA = () => ReactGA.initialize(process.env.GA_TRACKING_ID);
+export const trackCurrentPage = () => ReactGA.pageview(window.location.pathname);
```
nitish24p: Ideally this track page should be called only if GA has been initialised

src/containers/App/App.js
```diff
@@ -3,30 +3,37 @@ import { func } from 'prop-types';
...
-        >
-          {t('landing-page-start')}
-        </a>
-      </header>
-    </div>
-  );
+class App extends React.Component {
+  componentDidMount() {
+    initializeGA();
+    trackCurrentPage();
```
nitish24p: Both These calls can be part of a separate file like analytics.js..
suppose tomorrow we want to include mixpanel, or heap or some other lib, we add that in the abstracted analytics.js and it gets applied everywhere where ever analytics.js has been invoked

src/googleAnalytics.js
edward-hong: Yes I agree. I've read through react-ga docs again. There's no mention of a method that checks to see if GA is initialized. I guess instead of export const each function like I've done, I could export an object with these two functions as methods and a property called something like 'hasGABeenInit' which is a boolean and everytime the init function is called, 'hasGABeenInit' will then be set to true. Would that be the best approach?

src/containers/App/App.js
edward-hong: You've suggested to abstract the Google Analytics code. I thought that was a good idea so I abstracted it into a file called `googleAnalytics.js` in the `src` directory. I named the file name specifically enough to make sure that the code is for that one purpose. I'm not sure how the possibility of people wanting to add other libraries into the file that I made called `googleAnalytics.js` applies here.

scripts/custom-config.js
edward-hong: Thanks for the heads up! I'll make sure to make the adjustments you suggested.

src/googleAnalytics.js
edward-hong: @nitish24p I've changed the code in `googleAnalytics.js` is this better?

```
export default {
  hasGABeenInit: false,
  initGA: () => {
    ReactGA.initialize(process.env.REACT_APP_GA_TRACKING_ID);
    this.hasGABeenInit = true;
  },
  trackCurrentPage: () => {
    if (this.hasGABeenInit) {
      ReactGA.pageview(window.location.pathname);
    }
  }
}
```

nitish24p: The `this` context here will be the global window object not the object.. The idea is good but remove all the arrow functions for object method declarations. You can read more here https://dmitripavlutin.com/when-not-to-use-arrow-functions-in-javascript/
```javascript
export default {
  hasGABeenInit: false,
  initGA: function() {
    ReactGA.initialize(process.env.REACT_APP_GA_TRACKING_ID);
    this.hasGABeenInit = true;
  },
  trackCurrentPage: function() {
    if (this.hasGABeenInit) {
      ReactGA.pageview(window.location.pathname);
    }
  }
}
```

src/googleAnalytics.js
```diff
@@ -0,0 +1,14 @@
+import ReactGA from 'react-ga';
+
+export default {
+  hasGABeenInit: false,
+  initGA: function() {
```
arthurbuhl: I think we can use a shorter syntax for method definitions on objects initializers. Given the following code:

```js
var obj = {
  foo: function() {
    /* code */
  },
  bar: function() {
    /* code */
  }
};
```

You are now able to shorten this to:
```js
var obj = {
  foo() {
    /* code */
  },
  bar() {
    /* code */
  }
};
```

MDN Doc Link: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

README.md
```diff
@@ -72,3 +72,9 @@ You might have some conflicts while rebasing, try to resolve the conflicts for e
 In order to merge a PR, it will first go through a review process. Once it is approved, we will merge to the `development` branch using the `Squash` button in github.
 
 When using squash, all the commits will be squashed into one. The idea is to merge features/fixes as oppose of merging each individual commit. This helps when looking back in time for changes in the code base, and if the PR has a great comment, it's easier to know why that code was introduced.
+
+### Google Analytics
+
+Create a `.env` file in the root directory. Add your Tracking ID under the variable `REACT_APP_GA_TRACKING_ID` e.g:
+
+    REACT_APP_GA_TRACKING_ID='UA-xxxxxxxxx-x'
```
arthurbuhl: I think we should add a sample `.env` like `.env.sample` with all the ids and commit to git but with empty or demo values. This is done so that all the devs are in sync about all the env used in the app. Or maybe we can create a separate markdown file mentioning about all the env variables (just the varibale names, not the key values) used in the app.

src/containers/App/App.js
```diff
@@ -3,34 +3,37 @@ import { func } from 'prop-types';
...
-  );
+class App extends React.Component {
+  componentDidMount() {
+    GA.initGA();
+    GA.trackCurrentPage();
+  }
+  render() {
+    return (
+      <div className={styles.app}>
+        <img src={require('./assets/mentor.jpg')} alt="Mentor" className={styles.image} />
```
arthurbuhl: I think we should use `i18n` for this static text used for image alt.

README.md
edward-hong: Yea I think it's good for contributors to have some default keys to quickly get up and running

src/containers/App/App.js
edward-hong: This isn't relevant to this PR regarding Google Analytics and setup of `.env` as a consequence of GA

arthurbuhl: Yes. I think I will create a new issue for it.

README.md
arecvlohe: I think having an `.env.example` is a good idea so all the devs know what env vars there are

edward-hong: Ok I'll add an `.env.example` file with a GA Tracking so everyone can get up and running smoothly.

src/googleAnalytics.js
edward-hong: @arecvlohe You're right. I'll update the PR for this.

src/googleAnalytics.js
```diff
@@ -0,0 +1,14 @@
+import ReactGA from 'react-ga';
```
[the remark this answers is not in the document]
edward-hong: I've moved the files to `src/utils/ga.js`

arthurbuhl:
Any updates on this PR? Should we close it or keep it open for few more weeks?

crysfel:
A lot of things have changed since this PR was open. Given that there's no activity here, I'm going to close it, as it will be easier to just install `react-ga` and do the setup than fixing all the conflicts xD

Issue #56, opened 2018-09-14
Mission section


crysfel:
As a visitor I want to see the mission for the project.

We want to reuse the same component we used for the About section, all we need to do is send the data and additional custom css classes using the `classes` property.

Assets will be provided once ready.

nitish24p:
dibs ?

crysfel:
@tureey given that you are working on the `Section` component, it makes sense to assign this task to you xD

Issue #55, opened 2018-09-14
About section


crysfel:
As a visitor I want to learn more about the project. 

Create a react component that takes the title, description, direction, classes and icon as properties. This component should render the title, description and icon in the given direction (left, right).

Text will be provided later on, for now you can use a placeholder text.

We are going to use this component to render the mission section as well.

Assets will be uploaded to this tasks once are ready.

tureey:
Since I have no permissions in *Zenhub* yet, I say here Im working on it. Because I have no permissions either to change the labels of the issue and make it `WorkInProgress`

crysfel:
@tureey I've merged the PR that enables css modules! 

> What about let this component accept a prop called `actions`? 

Something like this?

```
<Section actions={{login: someFunction, facebook: someOtherFuncion }} />
```

If so.... how are we going to add the action to an element inside the component? What about:

```
<Section title="......" description=".........">
  <Action onClick={customAction} />
  <Image onClick={customAction} />
</Section>
```
And then render the children property in a designated place inside the `Section` component?

> Is the icon going to be a component because it will be a svg, png or just a plain text with the url?

We can create components for each image, then pass the icon as a property to the section. If we do this, the images will be inside the final JS bundle, it will render super fast but the tradeoff is that the bundle will be bigger. However, we are going to split the code making the bundles smaller.

tureey:
@crysfel  I was thinking about having 2 components for the `About` for example:

1.  Section (the one has background-color and the`::before` `::after` properties
2.  The Flex component that accepts
2.1 Title
2.2 Description
2.3 Direction
2.4 Icon
2.5 Actions  (*)

(*) I was thinking to make it be an array of `<Button>` and they would be placed inside a wrapper that places it inside the `SectionIconed`. Of course, optional.

I don't like the SectionIconed name so much. Any proposal? This component would be the one renders the content in a certain direction 
```
<Section {Still have to think about this props}>
   <SectionIconed
        title='About'
        text={'lorem ipsum lorem ipsum lorem ipsum lorem ipsum'}
        direction='left'
        icon={<Icon src={lazyload?}/>}
        ?actions={[
                <Button onClick={this.handleClickExample1}>,
                <Button onClick={this.handleClickExample2}>,
        ]}
    />
</Section>
```

PR #72, opened 2018-09-22
[GLOBAL/SCSS] Add Font-size variables

tureey:
# Details
## Description
Add font-size variable so everyone needs to use only this sizes.

PR #73, opened 2018-09-22
Feature/home sections

tureey:
# Details

## Description
For this task I am working on a `HomeSection` Component that given a `title`, `text` and `children` will prints the section. It's going to be enough for building **About**, **Mission**

## Issue
https://github.com/Coding-Coach/coding-coach/issues/55
https://github.com/Coding-Coach/coding-coach/issues/56

#### QUESTIONS
1. I asked to @emmawedekind but, how do we call the colors?

2. ~To use modifiers in a private/local class through props, how do we do that? I've seen people using a library called `classnames` but is there a _"native way"_ with `css modules`?~

3. ~How do you guys build the grid of the app?~
~_I thought that maybe the `<main></main>` should contain a class that gives it a `width` and `max-width` depending on the `resolutions`. With that, we assure that all the content is contained in the grid._~

There is the possibility to have too a `util/generic class` called for example `.u-layout` that gives that properties that we were talking about the assure the content has a `max-width`

**I am excited to hear about your proposals!** 

@crysfel @emmawedekind


src/components/Section/Section.js
```diff
@@ -0,0 +1,57 @@
...
+  } else if (bgc === 'primary') {
+    return styles.isPrimary;
+  } else if (bgc === 'secondary') {
+    return styles.isSecondary;
+  } else if (bgc === 'tertiary') {
+    return styles.isTertiary;
+  } else {
+    return '';
+  }
+}
```
arthurbuhl: I think `switch` statement can be used here as mentioned in this [post](https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04s02.html).

tureey: I was thinking about, yes! Thank you!

src/components/Section/Section.js
```diff
@@ -0,0 +1,66 @@
...
+    case 'grey':
+      return styles.isGrey;
+    case 'primary':
+      return styles.isPrimary;
+    case 'secondary':
+      return styles.isSecondary;
+    case 'tertiary':
+      return styles.isTertiary;
+  }
+}
```
arthurbuhl: Maybe we can optimize this `getBackgroundColorClass` function like:

```js
function getBackgroundColorClass(bgc) {
  return bgc ? styles[`is${bgc.charAt(0).toUpperCase() + bgc.slice(1)}`] : '';
}
```

tureey: That was the thing I was asking about generating dynamic styles! Nice! I tried a few things and none worked! Thank you so much!

tureey:
Now, the basic thing is done. I still have to:

1. Know how to use the mediaquery mixin (I tried a couple of thing and none work)
2. Decide How do the `grid`
3. Solve the problem with the separator alignment depending on the direction
4. Solve the problem with the separator color depending on the backgroundColor of the section
5. **Use the correct sizes** for `margins`, `padding` and `font-sizes` since I can't see them in Figma and I don't have macOS to use Sketch
6. Do the `::before` and `::after` for **About** and **Mission**

emmawedekind:
1. Read this: https://css-tricks.com/snippets/css/media-queries-for-standard-devices/
2. You shouldn't need grid: use CSS Flexbox
3 &4. (I think) You can use CSS pseudo classes for nth-child to grab alternating children: https://www.w3schools.com/cssref//sel_nth-child.asp
5. Don't worry about exact numbers - estimate. But keep it consistent.  Font sizes should be in a PR and the typescale is in the document. I'm using IBM Carbon's scale.

src/components/Section/Section.scss
```diff
@@ -0,0 +1,56 @@
...
+}
+
+
+.hasPaddingHorizontal {
+    padding-left: 30px; 
+    padding-right: 30px;
+}
+
+.hasPaddingVertical {
+    padding-top: 45px;
```
emmawedekind: This can be shorthanded to padding: 45px 0px;

src/pages/Home/components/SectionContent.js
```diff
@@ -0,0 +1,46 @@
...
+    <p className={styles.description}>{description}</p>
+    {children}
+  </div>
+);
+
+SectionContent.propTypes = {
+  title: PropTypes.string.isRequired,
+  // It makes the title have the color: MainColor or Grey
+  isColored: PropTypes.bool,
+  textAlign: PropTypes.oneOf(['left', 'right']),
```
emmawedekind: Just use the nth-child(odd) to grab alternating children.

src/pages/Home/components/SectionContent.scss
```diff
@@ -0,0 +1,49 @@
...
+        height: 1px;
+        background: $color-grey-light;
+    }
+}
+
+.title {
+    margin: 0;
+    font-size: 2rem;
+    color: $color-grey-dark;
+    $font-family: $font-alternative;
```
emmawedekind: Should be `font-family` not `$font-family`

src/pages/Home/components/SectionContent.scss
```diff
@@ -0,0 +1,49 @@
...
+    text-transform: uppercase;
+}
+
+.description {
+    line-height: 1.75;
+    font-size: 14px;
```
emmawedekind: Use the `$font-x` variables in #72 when available.

src/pages/Home/components/SectionDirection.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './SectionDirection.scss';
+
+function getClassesByDirection({ direction }) {
+  if (direction === 'left') {
```
emmawedekind: `return direction === 'left' ? styles.hasLeftDirection : styles.hasRightDirection;` But again... try to use CSS pseudo-classes instead of a variable.

src/theme/base/_variables.scss
```diff
@@ -1,4 +1,12 @@
 $font-main: 'Open Sans', sans-serif;
+$font-alternative: 'Fjalla One', sans-serif;
 
-$color-main: #EB6E64;
-$color-white: #FFF;
+$color-primary: #32B88C;
+$color-secondary: #69D5B1;
+$color-tertiary: #E6F5F0;
+
+$color-white: white;
```
emmawedekind: Use hex codes for black and white please. With 6 digits.

src/components/Section/Section.scss
tureey: Well, not really. I don't want to edit the horizontal padding property value. Doing this we are able to do the 4 posibilities:
- None
- Only vertical
- Only horizontal
- Both

src/pages/Home/components/SectionDirection.js
tureey: Mhh.. If you mean to use the pseudo-classes to create the order like:
Since **About is first** and **Mission is second**
`:nth-child(1) ---> make this flex-direction: row`
`:nth-child(2) ---> make this flex-direction: row-reverse`

I don't agree with that... Because you are editing the behaviour of the component from another point of the application. What if now instead of:
```
<Home>
    <About/>
    <Mission/>
</Home>
```
We have:
```
<Home>
    <Mission/>
    <About/>
</Home>
```

The components would behave different from what we expect **because now:**
**Mision is** `nth-child(1)`
**About is** `nth-child(2)`

But, what we can do is to add a `classname` to each instance, but then, we would have to add a new modifier to each new instance. It means that our `.scss` could grow as more instances we have

emmawedekind: I don't understand why you're talking about changing the order. 
image.

About & contact have the same layout (text right, images left).  Mission has the opposite. 

Is there a reason we can't use nth-child(even) to display the about and contact pages right.... I don't want to have a billion class names added everywhere for simple layout issues which can be fixed with pseudo classes. Styling with nth-child has nothing to do with DOM structure; just display.

emmawedekind: Use the pseudo-classes to style different/alternating layouts. Do not use left/right classes on everything.

src/components/Section/Section.scss
emmawedekind: Okay... but if you only want to edit the vertical, padding: 45px 0px is exactly the same as padding-top: 45px and padding-bottom: 45px. It's just more succinct.

tureey: No it's not the same! If I am wrong please make me understand it hahah. 
 says explicitly that I want `45px` on **top** and **bottom** but **says nothing about the left and right**.
```
padding-top: 45px;
padding-bottom:45px
```
Instead, if I say
```padding: 45px 0;```
I am saying **I want 45px on top and bottom**.

If we do: 
```
.hasVerticalPadding {
    padding: 45px 0;
}

.hasHorizontalPadding {
    padding: 0 45px;
}
```
And now I add both classes to an element, because of the CSS Cascade **(I don't know in `css modules`)**. Our element will have `padding: 45px 0` because is the first style computed. We will not be able either to combine them so we are losing 2 possibilities to combine.

File: src/pages/Home/components/SectionContent.scss (hunk 6 above)
Comment (tureey) [reply src/pages/Home/components/SectionContent.scss]: Yes! That why I asked for a merge as soon as possible, so we could be using the variables :smile:

File: src/pages/Home/components/SectionDirection.js (hunk 7 above)
Comment (emmawedekind) [reply src/pages/Home/components/SectionDirection.js]: No, you shouldn't repeat it. Create a mixin or something that we can include. Or place this in a Sass partial which extends to all components.

**Comment by tureey** [comment]
Her everyone. I almost finished the tasks, these are the tasks left:
1. **Use svg** for the images instead of the mock *(since I never configurated this... It looks like we need a svg loader)*
2. Change the `mediaqueries @mixin` to make them **mobiles first**
2.1. **Remove** the `manual mediaqueries`
2.2. **Update styles** with the new `mediaqueries @mixin`
3. Merge PR https://github.com/Coding-Coach/coding-coach/pull/72 So I can use the variables 

Since I changed many things, check the PR carefully
@crysfel @emmawedekind @arthurbuhl

File: src/theme/base/_elements.scss [hunk 9]
```diff
@@ -20,18 +20,18 @@
...
       background-color: $color-white;
-      color: $color-main;
+      color: $color-primary;
       border: 2px solid $color-white;
 
       &:hover {
-        background-color: $color-main;
+        background-color: $color-primary;
         cursor: pointer;
         color: white;
```
arthurbuhl: This needs to be `$color-white` from PR #72

src/pages/Home/components/HomeSection.scss
```diff
@@ -0,0 +1,80 @@
...
+
+.description {
+    color: $color-grey-dark;
+    font-size: 14px;
+    font-family: $font-main;
+    line-height: 1.5;
+    margin: 0;
+
+    @media screen and (min-width: 1024px) {
+        font-size: 16px;
```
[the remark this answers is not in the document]
tureey: yes, I talked to @crysfel saying that we should update our mediaqueries due to they are not **mobile-first**

tureey:
Like said before I was traveling. Tomorrow I'll update and pull everything and fix all of this! Thhank you all for the feedback though :smile:

src/pages/Home/components/HomeSection.scss
```diff
@@ -0,0 +1,80 @@
...
+        height: 1px;
+        width: 200px;
+    }
+}
+
+.title {
+    color: $color-grey-dark;
+    text-transform: uppercase;
+    font-family: $font-alternative;
+    font-size: 2.5rem;
```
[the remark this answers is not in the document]
crysfel: Now is merged!

src/pages/Home/components/HomeSection.scss
```diff
@@ -0,0 +1,80 @@
...
+    margin: 0;
+    line-height: 1.25;
+    text-transform: uppercase;
+}
+
+.description {
+    color: $color-grey-dark;
+    font-size: 14px;
```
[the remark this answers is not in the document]
crysfel: #72 has been merged!

src/containers/App/assets/app.scss
```diff
@@ -27,14 +27,14 @@
...
   margin: 0;
   text-align: center;
 }
 
 .tagline {
   color: $color-white;
-  font-size: 2em;
+  font-size: $font-xl;
   font-weight: 100;
   margin: 10px 0px;
```
emmawedekind: Please be consistent with margins. You've defined them with `rem` above. Stick with one unit.

src/components/contact/assets/contact.scss
```diff
@@ -25,11 +27,11 @@ section.contact {
...
     text-align: right;
     font-size: $font-xxl;
-    color: $green500;
+    color: $color-secondary;
     font-weight: $font-weight-thin;
     text-transform: uppercase;
-    border-bottom: 1px solid $gray;
-    font-family: $font-fjalla-one;
+    border-bottom: 1px solid $color-grey-light;
+    font-family: $font-alternative;
```
emmawedekind: Please rename this to font-secondary.

src/theme/base/_variables.scss
```diff
@@ -1,15 +1,17 @@
 // Font Family
 $font-main: 'Open Sans', sans-serif;
```
emmawedekind: $font-primary

src/theme/base/_variables.scss
```diff
@@ -1,15 +1,17 @@
...
-$font-fjalla-one: 'Fjalla One', sans-serif;
+$font-alternative: 'Fjalla One', sans-serif;
```
emmawedekind: $font-secondary

src/containers/App/assets/app.scss
crysfel: @emmawedekind are we going to have consistent white space in the designs? If so we definitely would have to create some constants or common css classes to avoid this kind of issues.

src/theme/base/_utils.scss
```diff
@@ -0,0 +1,12 @@
+@mixin generalAlignment() {
```
crysfel: I'd prefer to use composition rather than a mixin.

A mixing will add the same code in different places, increasing the final bundle. If instead you compose a class, it will use the same class everywhere.

```
.generalAlignment {
  margin-left: auto;
    margin-right: auto;
    
    @media screen and (min-width: 768px) {
        max-width: 992px;
    }
     @media screen and (min-width: 1400px) {
        max-width: 1200px;
    }
}
```

The, where you want to use it:

```
@media screen and (min-width: 768px) {
  . sectionInner {
    composes: generalAlignment;
    // more styles for sectionInner
  }
}
```

This will result in reusing the `.generalAlignment` in the DOM element. This is automatically done by webpack ;)

```
<div class="sectionInner generalAlignment" >
```

emmawedekind: Great idea!

tureey: Great idea! I did not know about the bundle size. Good to know thank you :smile:

tureey:
This afternoon/evening will leave this finished with all the conflicts solved and all the landing page finished.

tureey:
Yes. Thank you!

emmawedekind:
image
I am getting this error in the console... please fix.

emmawedekind:
Additionally, the social media link hover and focus state is broken. Please make the SVG of the icon white on hover/focus. Additionally, the transition takes too long. Please use `transition: background-color .1s linear` for now. Same for the icon fill.

tureey:
@emmawedekind Looks like since the guy who work on that did the refactor from `snake_case` to `camelCase` was not working. Everything looks fine now. 

Thank you everyone for all the feedback!

Well done guys :+1: :smile:

src/pages/home/Home.js
```diff
@@ -1,28 +1,54 @@
...
+              <span>{t('Coding')}</span>
+              <br />
+              <span>{t('Coach')}</span>
+            </h1>
+            <h2 className={styles.heroSubtitle}>
+              {t('Connecting developers with mentors worldwide')}
+            </h2>
+            <button className={styles.callToAction}>{t('Learn More')}</button>
+          </header>
+          <img className={styles.heroImage} alt="coding coach logo" src={heroImage} />
```
palashmon: I think we should use i18n for this static alt text also, just like we have done in all other places.

src/pages/home/Home.scss
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    &:before {
+        content: '';
+        position: absolute;
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
+        background: url('./assets/images/main-header-background.png') no-repeat;
```
palashmon: I think we should use SVG file here, just like we are doing it for all other images like `about.svg`, `mission.svg`, `meeting.svg`, etc.

src/pages/home/Home.js
tureey: I actually translated them... They have been overwritten by a merge or something!
**Good point** :+1:

src/pages/home/Home.scss
crysfel: There's a PR already for this, just leave this as it is.

emmawedekind:
Approved!

tureey:
last push! @crysfel @emmawedekind 
There were conflicts again. **I've solved them** :tada:

Issue #68, opened 2018-09-20
Package.json homepage


palashmon:
https://github.com/Coding-Coach/coding-coach/blob/0bad77aa8fd74608961ed9a90c1b19303775bd41/package.json#L5

This home page set in package.json is no loading and giving a 404 error. I think it should be [codingcoach.io](https://codingcoach.io/) instead. Please check this once.

Regards!

edward-hong:
@palashmon thanks for pointing this out! I'll make a quick PR regarding this issue.

emmawedekind:
I've approved. Merge in to develop once someone else approves.

crysfel:
Closing issue as the PR has been merged!

PR #74, opened 2018-09-23
fix home field in package.json

edward-hong:
## Description

Currently the home field in `package.json` points to http://coding-coach.github.io/ which points to 404 page. I've made the change to https://codingcoach.io/

## Issue

#68 Package.json homepage

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "coding-coach",
   "version": "0.1.0",
   "private": true,
-  "homepage": "http://coding-coach.github.io/",
+  "homepage": "https://codingcoach.io/",
   "dependencies": {
     "classnames": "^2.2.6",
     "i18next": "^11.7.0",


arthurbuhl:
Hi @emmawedekind, if I approve this would it help or only devs with special permission only can approve a PR?

emmawedekind:
Go ahead and try! @arthurbuhl

emmawedekind:
Hmm I don't know if it will work.... people don't have write access though so...

emmawedekind:
Ah I think @crysfel has to approve :/

arthurbuhl:
Yeah, I think so :unamused:

emmawedekind:
Ah dang, well thank you! I'm not sure why that is...

arthurbuhl:
No probs. I think that is a good system implemented by GitHub in a way, as that allows only valid reviewers suggestions are taken into consideration and no one else can just come in, out of no where and just start approving all the PRs without even having the knowledge of what's its all about. Making the process more meaningful and trustworthy.

PR #78, opened 2018-09-25
fix: use correct case sensitiv of socialMedia.scss file

DanielRuf:
# Details

## Description

This PR fixes the case sensitivity of the reference of socialMedia.scss which caused a broken build on Windows.

## Steps to QA

- [ ] run `yarn start` on Windows


arecvlohe:
How did this create a broken build? The correct name should be `socialmedia.scss`

DanielRuf:
> socialmedia.scss

Well, the file is named `socialMedia.scss` so it fails on Windows on `yarn start` as Windows is case sensitive.

I just changed the import as the filename casing (camelCase) might be connected to the actual component name (I'm not 10% sure but this makes some sense).

arecvlohe:
The original file should be `socialmedia.scss` so the correction should be with the filename not the import

DanielRuf:
> The original file should be `socialmedia.scss` so the correction should be with the filename not the import

Done. For those who wonder why a normal rename does not trigger a change for Git, try `git mv`.

[Reference](https://stackoverflow.com/questions/2641146/handling-file-renames-in-git)

arecvlohe:
@DanielRuf I have ran into this same issue a couple times when git does not pick up a filename change. It's kind of annoying!

PR #71, opened 2018-09-21
#54 Main Header

tarang9211:
# Details
PR for main-header component

## Issue
This fixes - https://github.com/Coding-Coach/coding-coach/issues/54

PR #80, opened 2018-09-25
fix: properly size image using object-fit: cover

DanielRuf:
# Details

## Description

Currently the image is skewed depending on the viewport. This partly fixes it. A better alternative solution would be to use `background-image` + `background-size: cover`.

## Steps to QA

- [ ] resize the viewport and check the image

### DIFF
--- a/src/containers/App/assets/app.scss
+++ b/src/containers/App/assets/app.scss
@@ -14,6 +14,7 @@
   left: 0;
   z-index: -1;
   min-width: 900px;
+  object-fit: cover;
 }
 
 .header {


arecvlohe:
@DanielRuf I was thinking the same thing. Should we just go ahead and update the image using your alternative solution? Seems to be more ideal

crysfel:
There's an open PR that removes all the code in this page, I don't think is worth investing more time on this, do you agree?

DanielRuf:
@crysfel which PR? I'm happy to review it and close mine.

crysfel:
@DanielRuf this one: https://github.com/Coding-Coach/coding-coach/pull/71/files#diff-0556f03522b596f0e507920a4e2fae60L10

DanielRuf:
Ok, I'm closing mine then. Stacking backgrounds is better in general.

PR #85, opened 2018-09-25
feature: use selfhosted fonts

DanielRuf:
# Details

## Description

self hosting fonts is typically faster and we can prevent FOIT and other disadvantages of externally loaded / hosted fonts.

## Steps to QA

- [ ] do an audit (Lighthouse against this PR and current `development`)

## Screenshots
with self hosted fonts and adding the CSS to the CRP (critical CSS):
stats10

current status (FOIT and higher loading times):
stats20
stats21


DanielRuf:
For further information, I did some more testing here and all our projects use this setup now for a better perceived performance and prevent reflows and other problems:

https://github.com/DanielRuf/my-slides/blob/master/00-webfonts/slides.pdf

arecvlohe:
@DanielRuf We are using google fonts I think. Can those be self-hosted? What alternatives can be self-hosted with minimal cost?

arecvlohe:
I see that Lato is free. Probably should include the license.

arecvlohe:
Lato license
[OFL.txt](https://github.com/Coding-Coach/coding-coach/files/2417329/OFL.txt)
[README-WEB.txt](https://github.com/Coding-Coach/coding-coach/files/2417330/README-WEB.txt)

arecvlohe:
Fjala license:
[SIL Open Font License.txt](https://github.com/Coding-Coach/coding-coach/files/2417333/SIL.Open.Font.License.txt)

DanielRuf:
> @DanielRuf We are using google fonts I think. Can those be self-hosted? What alternatives can be self-hosted with minimal cost?

Yes, the tool is also recommended by Google.

DanielRuf:
https://github.com/google/fonts
https://github.com/google/fonts#self-host-fonts-available-from-google-fonts

The licenses are not needed here as we do not ship the sources afaik.

DanielRuf:
Also you do not get them with the downloads ;-) Otherwise Google would not recommend the tool.

arecvlohe:
@DanielRuf If it's a third party resource should we not always include the license? I could see if we are getting it from google fonts, I suppose they host the license. But if we have it locally are we not suppose to have the license ourselves? That's how it is with `node_modules`

DanielRuf:
https://opensource.stackexchange.com/questions/7035/where-include-sil-open-font-license-in-a-website-if-i-use-font-face

And there is still the metdata normally.

Sae with Gatsby and typography.js and other big projects.
https://github.com/KyleAMathews/typefaces/blob/master/scripts/download-all-google-fonts.js

DanielRuf:
Still, the Google APIs do not provide them by default and it is not required to include them.

DanielRuf:
The right licenses would be

https://github.com/google/fonts/blob/master/ofl/lato/OFL.txt
https://github.com/google/fonts/blob/master/ofl/sourcesanspro/OFL.txt
https://github.com/google/fonts/blob/master/ofl/fjallaone/OFL.txt

Also see https://theleagueof.github.io/licenses/ofl-faq.html

DanielRuf:
> 1.10 Does the full OFL license text always need to accompany the font?
> The only situation in which an OFL font can be distributed without the text of the OFL (either in a separate file or in font metadata), is when a font is embedded in a document or bundled within a program. In the case of metadata included within a font, it is legally sufficient to include only a link to the text of the OFL on http://scripts.sil.org/OFL, but we strongly recommend against this. Most modern font formats include metadata fields that will accept the full OFL text, and full inclusion increases the likelihood that users will understand and properly apply the license.

DanielRuf:
image

DanielRuf:
https://fontdrop.info/
https://opentype.js.org/font-inspector.html

Because the OFL license files are never transferred to the user(s) in this case but they are already part of the metadata of the files.

arecvlohe:
I think this means having the license accompany the font is encouraged but I am not a lawyer either haha

>1.11 What do you mean by 'embedding'? How does that differ from other means of distribution?
By 'embedding' we mean inclusion of the font in a document or file in a way that makes extraction (and redistribution) difficult or clearly discouraged. In many cases the names of embedded fonts  might also not be obvious to those reading the document, the font data format might be altered, and only a subset of the font - only the glyphs required for the text - might be included. Any other means of delivering a font to another person is considered 'distribution', and needs to be accompanied by any copyright notices and licensing information available in OFL.txt.

src/theme/base/_fonts.scss
```diff
@@ -0,0 +1,26 @@
+/* open-sans-regular - cyrillic-ext_vietnamese_cyrillic_greek_greek-ext_latin_latin-ext */
```
crysfel: If the fonts are already defined in the `public/index.html` file, is it necessary to have this file?

src/theme/index.scss
```diff
@@ -1,6 +1,7 @@
 @import "base/variables";
 @import "base/elements";
 @import "base/mediaqueries";
+//@import "base/fonts";
```
crysfel: the fonts file is not being imported here. In terms of performance, having the fonts inline (in the html file) makes sense to me, but in terms of code standards I like the idea of having a separate file and import it here.

src/theme/base/_fonts.scss
DanielRuf: It is currently there if you need this instead of the "critical CSS" approach to prevent FOIT and FOUT.

But it is not actively used as the import to this file is commented out.

src/theme/index.scss
DanielRuf: See the results of my benchmarks, critical CSS prevents FOIT and FOUT and is often better regarding performance (results). This line here and the file are open for discussion and are the alternative approach (but trigger FOIT and worse performance).

crysfel: I'd say we go for performance, thoughts?

Issue #81, opened 2018-09-25
process.env


DanielRuf:
# Issue

## Description

Using `process.env` is not great as the used env variables seem to be not declared anywhere. We should use `cross-env` for testing purposes and use some config for setting them instead of using `process.env`.

linux-nerd:
@DanielRuf Thanks for pointing it out. I added .env files with the variables as part of #57, and I think I missed to commit it.
I will raise a PR against this issue.

linux-nerd:
Ahh! Now I see why I missed to add .env file. @crysfel, @emmawedekind I think we can remove .env file from .gitignore. It makes sense to ignore .env.local or.env.* but not .env file. If the developer has to change the value of .env file, hr/she can create .env.local file and make changes. Thoughts?

emmawedekind:
I think that's fine! I have approved the PR.

edward-hong:
When I started the Google Analytics PR #70, I also included setting up `.env` for the tracking ID. It might have been a good idea for me to have started a separate issue. I'll leave the `.env` stuff to @linux-nerd and focus on getting the GA PR done.

crysfel:
I'm wondering if `dotenv` supports `.env.local` to overwrite the values from `.env`. As far as I know CRA internally uses dotenv to load data to the `process.env` object.

charlesjliu:
I don't think we should have a `.env` committed into the repo. 

https://www.npmjs.com/package/dotenv#should-i-commit-my-env-file

Is there a reason why those links are set using environment variables (slack etc.)? I don't think those should be set by the environment.

If anything, you could set default values if the environment variables aren't detected. We can have a sample `.env` file that developers can use to write their own and include instructions in the repo. But environment variables should be for things like... the `CODING_COACH_API` url, or whatever client API keys we're using (like the Google Analytics example).

Just my two cents! 🤷‍♂️

DanielRuf:
Exactly. dot-env are normally meant for credentials and user specific settings.

PR #88, opened 2018-09-26
Fix: Removed .env file entry from .gitignore and added .env file

linux-nerd:
# Details

## Description

- Removed .env file entry from .gitignore
- Added .env file

## Issue

Issue opened by @DanielRuf #81


arecvlohe:
@linux-nerd I think the plan is to have a `.env.example` file and to keep `.env` hidden

arecvlohe:
@linux-nerd #70 in reference to the `.env` stuff

crysfel:
An `.env` file is usually unique for each environment, we don't want this in the code base, instead (as @arecvlohe mentioned) we are going to have a `.env.example` that we can use as a reference.

PR #96, opened 2018-09-29
Removing old landing page content

emmawedekind:
Removing the old landing page content since it will be overwritten by the new components.


emmawedekind:
Ahh okay I didn't realize #71 took care of this.

Issue #99, opened 2018-10-01
Main Header Follow-Up: Remove additional `<h1>`


crysfel:
# Issue

## Description

In the home page, the current implementation uses two `h1`, we only need to have one `h1` in the entire page. 

The scope of this task limits to remove the second `h1` and put all the text in a single `h1` element, make sure to use the i18n library that we have.

linux-nerd:
@crysfel I will be working on this

tureey:
I did the refactor of a different things on my PR even before they decided to creat the `tech debt tasks`
Fixed here https://github.com/Coding-Coach/coding-coach/pull/73

crysfel:
PR merged, thanks!

PR #113, opened 2018-10-02
Fix: Main Header Follow-Up: Remove additional `<h1>`

linux-nerd:
# Details

## Description

- [x] Removed additional h1 tags
- [x] used i18n for text 

## Issue
#99


src/pages/home/assets/home.scss
```diff
@@ -28,13 +28,11 @@
...
 
-    &:nth-child(1) {
-      margin-bottom: 0px;
-    }
-
-    &:nth-child(2) {
-      margin-top: 0px;
-    }
+    // Break words in multiple lines
+    width: min-intrinsic;
```
crysfel: I like this!

Issue #98, opened 2018-09-30
Main Header Follow-Up: Remove png background and use css only 


emmawedekind:
# Issue
Relates to the Main Header Epic #54

## Description
The current landing page contains an extra div to set the background. While this works is not ideal, this task is about removing the following `div` from the landing page:

```
<div class="src-pages-home-home-mainHeaderBackground"></div>
```

Then we need to implement the background using css only, I've prepared an example on how to achieve this with only css: https://codepen.io/crysfel/pen/gdVyjW

Make sure to remove the `main-header-background.png` file from the assets folder.

reeversedev:
I'll be interested in doing this.

reeversedev:
I would be interested in doing this.

crysfel:
@reeversedev I will create smaller tasks for this, so other people can take something

AlexMunoz:
@emmawedekind can you review it :)

Issue #100, opened 2018-10-01
Main Header Follow-Up: Use SVG instead of PNG


crysfel:
# Issue

## ZenHub Issue

The home page is displaying an illustration in the right side, currently is a `assets/logo.png` image, however we need to use a SVG instead of a png file. The reason for this is because we want to display nice and crispy images in retina/large displays.

The scope of this task is about removing the png file and using an svg file instead.

linux-nerd:
@crysfel I can take this up.

PR #115, opened 2018-10-02
Main Header: Background using css

AlexMunoz:
# Details

## Description

remove the div and .png file and implement the background using css. 

## Issue

- #98


FacundoGFlores:
Good job. I would rather suggest to move the colors to `_variables.scss`. The palette should define a `$white` and also the greens in the background as well.

src/pages/home/assets/home.scss
```diff
@@ -1,16 +1,9 @@
...
-
-.mainHeaderBackground {
-  background: url('./main-header-background.png') no-repeat;
-  height: 100%;
-  width: 100%;
-  background-size: cover;
-  position: absolute;
-  z-index: -1; 
+  min-height: 100vh;
+  background-color: #6dd4b1;
```
crysfel: Please use the colors defined in `src/theme/base/variables.scss`.

emmawedekind:
I'm sorry but I'm confused by all the comments on these issues.. can you please explain?

AlexMunoz:
i mean, #73 made better changes in the home layout to follow the design, this solution fix a the previous  state. I think it's better to apply #73 and then fix the issue with the background.

palashmon:
I think we should use SVG here also instead of PNG for `/assets/main-header-background.png` based on same critria mentioned in issue #100


> The home page is displaying an illustration in the right side, currently is a `assets/logo.png` image, however we need to use a SVG instead of a png file. The reason for this is because we want to display nice and crispy images in retina/large displays.

tureey:
Good point @AlexMunoz ! I think the same.

crysfel:
@AlexMunoz can you resolve the conflicts and update the code based on the comments? PR #73 has been merged!

AlexMunoz:
i fixed my commit

Issue #105, opened 2018-10-01
Main menu for mobile


crysfel:
# Issue

## ZenHub Issue

In mobiles we need to show a hamburger menu at the top-left instead of displaying the main menu (for desktops/tablets).

When the user clicks the hamburger icon, the menu should slide in with a smooth animation. Internally it should have a close button, and three links: About, Mission, Contact.

Please follow the specs: https://www.figma.com/file/MfEpHTzcJZTnclKSojQQdhpp/CodingCoach-Design-System?node-id=529%3A62

dhrubesh:
Hey, I will take this.
But just wanted to confirm if the `Button` component is ready or not? If not, I can work on it as well.

crysfel:
I've assigned the button component to you: https://github.com/Coding-Coach/coding-coach/issues/112

dhrubesh:
Sure Thanks! Can you assign this issue to me as well?

tureey:
Since we will not do this I close the issue

PR #129, opened 2018-10-05
Feature/main menu

dhrubesh:
# Details

## Description

Added Main Menu for Mobile.

## Issue
#105

### DIFF
--- a/src/components/navbar/Navbar.js
+++ b/src/components/navbar/Navbar.js
@@ -3,20 +3,26 @@ import { translate } from 'react-i18next';
 import 'i18n/i18n';
 import styles from './navbar.scss';
 
-const Navbar = ({ t }) => (
-  <nav className={styles.navigationBar}>
-    <div className={styles.inner}>
-      <a href="#about" className={styles.navLinks}>
-        {t('About')}
-      </a>
-      <a href="#mission" className={styles.navLinks}>
-        {t('Mission')}
-      </a>
-      <a href="#contact" className={styles.navLinks}>
-        {t('Contact')}
-      </a>
-    </div>
-  </nav>
-);
+const Navbar = ({ t }) => {
+  const renderDesktop = (
+    <nav className={styles.navigationBar}>
+      <div className={styles.inner}>
+        <a href="#about" className={styles.navLinks}>
+          {t('About')}
+        </a>
+        <a href="#mission" className={styles.navLinks}>
+          {t('Mission')}
+        </a>
+        <a href="#contact" className={styles.navLinks}>
+          {t('Contact')}
+        </a>
+      </div>
+    </nav>
+  );
+  const renderMobile = <div>This is for mobile.</div>;
+  const renderComponent = window.innerWidth > 600 ? renderDesktop : renderMobile;
+
+  return <div>{renderComponent}</div>;
+};
 
 export default translate('translations')(Navbar);
--- a/src/components/navbar/Navbar.js
+++ b/src/components/navbar/Navbar.js
@@ -3,7 +3,7 @@ import { translate } from 'react-i18next';
 import 'i18n/i18n';
 import { slide as Menu } from 'react-burger-menu';
 import styles from './navbar.scss';
-
+import './burgerMenu.scss';
 const Navbar = ({ t }) => {
   const BurgerMenustyles = {
     bmBurgerButton: {
@@ -60,7 +60,7 @@ const Navbar = ({ t }) => {
   );
   const renderMobile = (
     <div>
-      <Menu styles={BurgerMenustyles} isOpen={false}>
+      <Menu isOpen={false}>
         <a href="#about" className={styles.navLinks}>
           {t('About')}
         </a>
--- a/src/components/navbar/Navbar.js
+++ b/src/components/navbar/Navbar.js
@@ -1,9 +1,48 @@
 import React from 'react';
 import { translate } from 'react-i18next';
 import 'i18n/i18n';
+import { slide as Menu } from 'react-burger-menu';
 import styles from './navbar.scss';
 
 const Navbar = ({ t }) => {
+  const BurgerMenustyles = {
+    bmBurgerButton: {
+      position: 'fixed',
+      width: '36px',
+      height: '30px',
+      left: '36px',
+      top: '36px',
+    },
+    bmBurgerBars: {
+      background: '#21785b',
+    },
+    bmCrossButton: {
+      height: '24px',
+      width: '24px',
+    },
+    bmCross: {
+      background: '#979797',
+    },
+    bmMenu: {
+      background: '#afe9d6',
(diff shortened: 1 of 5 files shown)


dhrubesh:
In this PR I'm not handling `Rerender view on browser resize` . I would appreciate if @crysfel or @emmawedekind can create an issue on the same and assign it to me. I will create a HOC for the same.

tureey:
well, +1 to what Emma said

dhrubesh:
Hey @emmawedekind apparently importing the external CSS isn't working in this case and hence I'm unable to use the **variable** color. Can I create another file as _constant.js_ which will hold the same variables as _/src/theme/base/_variables.scss_ ? So that it could be used by components which expect styles to be passed as **JS objects**.

src/components/navbar/Navbar.js
```diff
@@ -1,22 +1,81 @@
...
-        {t('About')}
-      </a>
-      <a href="#mission" className={styles.navLinks}>
-        {t('Mission')}
-      </a>
-      <a href="#contact" className={styles.navLinks}>
-        {t('Contact')}
-      </a>
+const Navbar = ({ t }) => {
+  const BurgerMenustyles = {
```
crysfel: Can we move these styles to a `scss` file?

src/components/navbar/Navbar.js
```diff
@@ -1,22 +1,81 @@
...
+          {t('Mission')}
+        </a>
+        <a href="#contact" className={styles.navLinks}>
+          {t('Contact')}
+        </a>
+      </div>
+    </nav>
+  );
+  const renderMobile = (
+    <div>
```
crysfel: Looks to me that this `div` can be removed.

src/components/navbar/Navbar.js
```diff
@@ -1,22 +1,81 @@
...
+      <Menu styles={BurgerMenustyles} isOpen={false}>
+        <a href="#about" className={styles.navLinks}>
```
crysfel: Can you eliminate the redundancy? just put the links in a variable and use it in both places.

src/components/navbar/Navbar.js
```diff
@@ -1,22 +1,81 @@
...
+          {t('Contact')}
+        </a>
+      </Menu>
     </div>
-  </nav>
-);
+  );
+  const renderComponent = window.innerWidth > 600 ? renderDesktop : renderMobile;
+
+  return <div>{renderComponent}</div>;
```
crysfel: the extra div here is not needed, please remove it.

src/components/navbar/Navbar.js
```diff
@@ -1,22 +1,81 @@
...
+        </a>
+        <a href="#contact" className={styles.navLinks}>
+          {t('Contact')}
+        </a>
+      </Menu>
     </div>
-  </nav>
-);
+  );
+  const renderComponent = window.innerWidth > 600 ? renderDesktop : renderMobile;
```
crysfel: is it possible to use media queries for this? Having this logic in javascript is not ideal.

src/components/navbar/navbar.scss
```diff
@@ -30,4 +30,9 @@
   &:not(:last-child) {
     margin-right: 2em;
   }
-}
\ No newline at end of file
+}
+@media screen and (max-width: 600px) {
```
crysfel: Please use the `respond-to` https://github.com/Coding-Coach/coding-coach/blob/development/src/theme/base/_mediaqueries.scss#L36

src/components/navbar/Navbar.js
dhrubesh: I tried that already but due to some reason, the _Menu_ component isn't appending classes to the required child. Hence decided to keep it as it is.

src/components/navbar/Navbar.js
dhrubesh: I tried moving it to _scss_ then importing back but apparently, for some reason, the styles aren't not getting applied. I'll try updating this branch tomorrow.

src/components/navbar/Navbar.js
dhrubesh: Are you talking about something of this sort: _https://www.npmjs.com/package/react-responsive_

emmawedekind:
Please do not create an additional, duplicate file. Why can't you use variables.scss?

dhrubesh:
@emmawedekind @crysfel Can you guys check once why _.scss_ file isn't loading?

src/components/navbar/Navbar.js
crysfel: Is not loading because you are not using them in the react component, consider doing something like this:

```
import BurgerMenuStyles from './burgermenu.scss';
```

We are using css modules, therefore the final bundle only includes what we use inside the react components, by importing the styles and adding the classnames in an object then using it where you need, webpack will include them in the final bundle

Please also use lowercase for the scss files.

src/components/navbar/burgerMenu.scss
```diff
@@ -0,0 +1,52 @@
+/* Position and sizing of burger button */
+.bm-burger-button {
```
crysfel: Please use camel case for the class names, the reason for this is to being able to access these styles from JavaScript easier

src/components/navbar/Navbar.js
dhrubesh: Hey, I tried that as well.
If I do it the following way:
`import burgerStyles from './burgerMenu.scss';`
and 
`console.log(burgerStyles)`
Its throwing the following output: 
image

whereas we except it to be the following way:
`{
    bmBurgerButton: {
      position: 'fixed',
      width: '36px',
      height: '30px',
      left: '36px',
      top: '36px',
    },
    bmBurgerBars: {
      background: '#21785b',
    },
    bmCrossButton: {
      height: '24px',
      width: '24px',
    },
    bmCross: {
      background: '#979797',
    },
    bmMenu: {
      background: '#afe9d6',
      padding: '2.5em 1.5em 0',
      fontSize: '1.15em',
    },
    bmMorphShape: {
      fill: '#373a47',
    },
    bmItemList: {
      color: '#b8b7ad',
      padding: '0.8em',
    },
    bmItem: {
      display: 'block',
      padding: '1.2em',
    },
    bmOverlay: {
      background: 'rgba(0, 0, 0, 0.3)',
    },
  };`

That's the reason I'm stuck.

src/components/navbar/burgerMenu.scss
dhrubesh: These class-names are automatically generated by `react-burger-menu`.

crysfel: then you will have to wrap them inside `:global { ..... }`, which is not ideal but needed in order to integrate libraries that are not using css modules.

crysfel:
Given that the new design doesn't need a hamburger menu, I'm going to close this PR. Maybe we can use something like this later down the road

PR #135, opened 2018-10-07
Fix: 'Coding Coach' text is coming in a single line. 

linux-nerd:
# Details

## Description

- [x] Changed 'Coding\nCoach' to 'Coding Coach'
- [x] Added styles from #113 to break the words in multiple lines
- [x] Removed unused assets/home.scss file
- [x] Renamed Home.scss to home.scss and moved it inside assets directory.
- [x] Changed `@media screen and (min-width: 768px)` to `@include respond-to(pc)`
- [x] Added responsive handler for screen size 1024 `@media screen and (min-width: 768px)` - > 
```
 $break-xlarge: 1024px;
------
 @else if $media==largepc {
    @media (min-width: $break-xlarge) {
        @content;
    }
  }
```
- [x] Changed rem to px for padding and margins, considering default pixel size of 16px
## Issue

#128

Issue #134, opened 2018-10-07
CI tests failing


arthurbuhl:
# Issue

## Description

Tests are failing on travis CI
Link - https://travis-ci.org/Coding-Coach/coding-coach/builds/438155969
Script - `npm run test`

## Failed Files

-  src/components/contact/__ tests __/SocialMedia.test.js
   -  `expect(value).toMatchSnapshot()` - Received value does not match stored snapshot 1.
-  src/components/contact/__ tests __/Contact.test.js
   - Unable to find an element with the alt text: people search logo 
- src/containers/App/__ tests __/App.test.js

emmawedekind:
I've restarted the build... hopefully it passes now.

arthurbuhl:
The build was actually terminated this time at it was running for more than 10 mins and a normal CI build usually takes about 2-3 mins:

```bash
Webpack Bundle Analyzer is started at http://127.0.0.1:8888
Use Ctrl+C to close it

No output has been received in the last 10m0s, this potentially indicates a stalled build 
or something wrong with the build itself.
Check the details on how to adjust your build configuration on: 
https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

I think `Webpack Bundle Analyzer` script should not be present in Travis CI build. That might be the issue for the stalled build.

linux-nerd:
@arthurbuhl, @emmawedekind The test for Contact.test.js failed because the alt text of logo was changed to 'Person using a magnifying glass to see the details of another guy.      ' and the one used in the test file is 'people search logo'

arthurbuhl:
Ok, also I found that translation is not used in Contact.test.js and text there is hard-coded in English there. Will that cause an issue failing the test in any scenario?

arthurbuhl:
Also, it would be great if we add test script to `lint-staged` so that before any commit we always make sure tests are running fine, something like 

https://github.com/facebook/create-react-app/issues/2336

emmawedekind:
I think the build is passing now... can someone verify this?

arthurbuhl:
@emmawedekind yes, CI is passing now. @crysfel just needs to approve the PR #136 and then you should be able to merge it.

PR #136, opened 2018-10-07
WIP: Fix: CI tests failing

linux-nerd:
# Details

## Description
I tried solving this issue in my PR #135, but it seems it requires some discussion. So, instead of blocking that PR I created a new one.

- [x] Removed dependency of .env file to point to correct urls (also being discussed in #81)
- [x] Updated the snapshots
- [ ] Unit test is failing for App.test.js. I have a strong feeling that the test is failing because of @react/router. Their documentation does not say anything about testing with reach router. I would appreciate any help in this regards.
 
## Issue
#134

### DIFF
--- a/src/components/contact/Contact.js
+++ b/src/components/contact/Contact.js
@@ -6,14 +6,18 @@ import styles from './assets/contact.scss';
 import peopleSearch from './assets/people-search.svg';
 import SocialMedia from './SocialMedia';
 
+export const EMAIL = 'codingcoachio@gmail.com';
+export const SLACK_URL =
+  'https://join.slack.com/t/coding-coach/shared_invite/enQtNDMyMzUzNjAxODQyLTUwYTFkMzZmNGFhMzZjYTQwOWE2YWFjOGJhNzYyOGIxNDM5Zjc1YWQ4NjMwN2U5YzBlNjkwMTI2ZDNiOWQyMzM';
+
 const Contact = ({ t }) => {
   const emailLink = (
-    <a target="_blank" href={`mailto:${process.env.REACT_APP_EMAIL}`} className={styles.email}>
-      {process.env.REACT_APP_EMAIL}
+    <a target="_blank" href={`mailto:${EMAIL}`} className={styles.email}>
+      {EMAIL}
     </a>
   );
   const slackOrg = (
-    <a href={process.env.REACT_APP_SLACK_URL} className={styles.slackOrg}>
+    <a href={SLACK_URL} className={styles.slackOrg}>
       {t('slack-org')}
     </a>
   );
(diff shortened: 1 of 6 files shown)


src/components/contact/Contact.js
```diff
@@ -6,14 +6,18 @@ import styles from './assets/contact.scss';
 import peopleSearch from './assets/people-search.svg';
 import SocialMedia from './SocialMedia';
 
+export const EMAIL = 'codingcoachio@gmail.com';
```
crysfel: Let's create a configuration file, where we get the value from the environment or assign a default value if empty: `src/config/Constants.js`

```
export default {
  contact: {
    EMAIL:  process.env.EMAIL ||  'codingcoachio@gmail.com',
    SLACK_URL: process.env.SLACK_URL ||  '......',
  },
  social: {
    FB_URL: process.env.FACEBOOK_URL || 'https://www.facebook.com/codingcoachio/',
    INSTA_URL: process.env.INSTA_URL ||  'https://www.instagram.com/',
    TWITTER_URL: process.env.TWITTER_URL  || 'https://twitter.com/codingcoach_io',
  },
};
```

Then we will use the config object with the values from the env or the default.

crysfel:
@linux-nerd I don't think is worth testing the router, after all it's already tested in their own repo. Let's remove that test.

crysfel:
It looks like everything went well in the build, but still is showing as failed. I think that's because the bundle analyzer is running an http server.

Maybe we can send a parameter in the CI to avoid running the analyzer there. I've seen that there's an option to add a custom script, we might be able to send a parameter to enable the analyzer whenever we would like to generate it, can you take a look?

https://github.com/Coding-Coach/coding-coach/blob/development/scripts/config-overrides.prod.js#L37

Maybe something like this:
```
if (process.argv[0] === 'analyze') {
  // Adds the bundle analyzer to inspect the production build
  config.plugins.push(new BundleAnalyzerPlugin());
}
```

Another option is to take a look at the parameters received by `BundleAnalyzerPlugin` and see if we can disable the http server and instead just saving the files in the disk.

linux-nerd:
sure, I will look at it tomorrow  :)

package.json
```diff
@@ -38,14 +39,16 @@
     "**/src/**/*.js": [
       "prettier --write",
       "eslint --fix",
-      "git add"
+      "git add",
+      "test"
```
arthurbuhl: This should be `test:staged` instead of just `test`, right?
Also, we should have test script before `git add`, so that we make sure that `git add` only happens if the tests passed successfully.

emmawedekind:
image
Seems like the build is passing now - do I need to do anything else?

Issue #127, opened 2018-10-05
Reusable Components: Panel


crysfel:
# Description
We need a `Panel` component that we can use anywhere in our application. It should be able to accept  a title, content and components in the footer.

```
const classes = {
  root: styles.classNameForRootElement,
  title: styles.classForTheTitle,
  close: styles.classForTheCloseButton,
  body: styles.classForTheBodyContainer,
  footer: 'custom-class-for-the-footer-container',
};

<Panel closable={false} onClose={onClose} classes={classes} modal={true}>
  <Panel.Header>Some title here</Panel.Header>
  <Panel.Content>
    <img src="something.png" />
    <div>
       <p>We should be able to add anything as the content<.p>
    </div>
  </Panel.Content>
  <Panel.Footer>
    <CustomFooterComponent />
  </Panel.Footer>
</Panel>
``` 

If the `modal` property is set to `false`, the panel should render in place. For example in a sidebar or anywhere else.

## Properties
- **title** A string with for the main title (Optional)
- **closable** If `true` the panel should display a close icon at the top, defaults to `false` (Optional)
- **onClose** A callback when use clicks the close icon (Optional)
- **modal** If `true` it shows the panel like a modal window, defaults to `false`
- **classes** An object with custom classes to apply to the inner elements of the panel (Optional)

## Design
![image](https://user-images.githubusercontent.com/7671983/46428177-987f0e80-c743-11e8-8e36-29f87b956771.png)

**The `Panel` component should only need the `title` and `background` styles from the design**, The content will be provided by others.

linux-nerd:
@crysfel I was thinking of something like this - 
```
const classes = {
  root: styles.classNameForRootElement,
  title: styles.classForTheTitle,
  close: styles.classForTheCloseButton,
  body: styles.classForTheBodyContainer,
  footer: 'custom-class-for-the-footer-container',
};

<Panel trigger={<Button>Open Me</Button>} closable={false} onClose={onClose} classes={classes}>
  <Panel.Header>Some title here</Panel.Header>
  <Panel.Content>
    <img src="something.png" />
    <div>
       <p>We should be able to add anything as the content<.p>
    </div>
  </Panel.Content>
  <Panel.Footer>
    <CustomFooterComponent />
  </Panel.Footer>
</Panel>
```

The trigger will mount the component passed to it and on click of that component the popup will be toggled. Although for this structure 3 extra components will be created but it looks neat. Any thoughts?

[...]

PR #123, opened 2018-10-04
Feature/mediaqueries mobilefirst

tureey:
# Details

## Description

Change `mediaqueries`to respond the **mobile first** methodology and update documentation/usage. After that, update the styles to do not break anything.

## Issue

https://github.com/Coding-Coach/coding-coach/issues/107

Issue #122, opened 2018-10-03
Footer


emmawedekind:
## Description
Implement the footer with links to respective modals.

![image.png](https://images.zenhubusercontent.com/5a3a7d108a75884b908bc2c0/ab3af15c-9408-4af5-87bd-61dee6695d3d)

Venablena:
I'll do this.

emmawedekind:
Assigned! Thanks!

Venablena:
I'm done implementing and styling the footer but since the modal isn't here yet, I haven't implemented links to it.  I could create a pull request now and create a new issue for adding the modal functionality or I can stand by and wait for it. OR I can go ahead and build the modal too but I believe @linux-nerd is already working on it.

crysfel:
In order to build the modals we need to create these three components first:

#112 
#125 <--- Would you like to take this one?
#127

emmawedekind:
Closing as this is implemented!

Issue #117, opened 2018-10-03
Privacy Policy


emmawedekind:
## Description

Privacy Policy MUST be created.  I can create a mock-up but we cannot deliver the landing page otherwise.

## Assets
[Here is a link](https://docs.google.com/document/d/19aG24XBouXTlRFnR37NHDIq2z7qCcFYMyVSQxmTz5dE/edit?usp=sharing) to the Google Doc with the Privacy Policy document.

## Design
![image](https://user-images.githubusercontent.com/7671983/46428177-987f0e80-c743-11e8-8e36-29f87b956771.png)

linux-nerd:
I can take this up :)

emmawedekind:
Super!

emmawedekind:
Please create this dialog so it's extensible for Terms & Conditions as well as Cookies.

crysfel:
I think we should break this task into thee more:

1.- We need a reusable `Button` component (We have a task for this: https://github.com/Coding-Coach/coding-coach/issues/112)
2.- A reusable `Panel` component with Title, Content and footer
3.- A reusable `TextField` component with `multiline` property to render a textarea.

Then we could create a `LegalDialog` component that looks exactly like this:

```
<LegalDialog title="Terms & Conditions" content="This terms and conditions...." onClose={doSomething}  />

<LegalDialog title="Private Policy" content="We the people..." onClose={doSomething}  />
```

linux-nerd:
@crysfel , @emmawedekind I am working on this. One question - When should this popup open? Open the click of ?

tureey:
In the footer there is a `footer-privacy` link. I guess that clicking there. @linux-nerd

crysfel:
@linux-nerd what @tureey said 👆

PR #139, opened 2018-10-08
Feature/footer

Venablena:
# Details

## Description
* Adds a footer to the landing page (without onClick behavior that opens the modal, pending issue #117) 
* Responsive behavior is currently not implemented as mobile-first (couldn't get it to work, pending merging of issue #123 ?)
* Modifies gray variables for text to better match design documentation
* TO DO: transform this into a reusable panel component (pending issue #127)

## Issue
#122

### DIFF
--- a/src/components/footer/Footer.js
+++ b/src/components/footer/Footer.js
@@ -1,0 +1,16 @@
+import React from 'react';
+import { translate } from 'react-i18next';
+import 'i18n/i18n';
+import styles from './footer.scss';
+
+const Footer = ({ t }) => (
+  <footer className={styles.footerMain}>
+    <div className={styles.footerInner}>
+      <a href="#terms">{t('footer-terms')}</a>
+      <a href="#cookies">{t('footer-cookies')}</a>
+      <a href="#privacy">{t('footer-privacy')}</a>
+    </div>
+  </footer>
+);
+
+export default translate('translations')(Footer);
--- a/src/components/footer/Footer.js
+++ b/src/components/footer/Footer.js
@@ -6,9 +6,15 @@ import styles from './footer.scss';
 const Footer = ({ t }) => (
   <footer className={styles.footerMain}>
     <div className={styles.footerInner}>
-      <a href="#terms">{t('footer-terms')}</a>
-      <a href="#cookies">{t('footer-cookies')}</a>
-      <a href="#privacy">{t('footer-privacy')}</a>
+      <a href="#terms" className={styles.footerLink}>
+        {t('footer-terms')}
+      </a>
+      <a href="#cookies" className={styles.footerLink}>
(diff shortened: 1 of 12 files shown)


src/components/footer/footer.scss
```diff
@@ -0,0 +1,34 @@
+@import '~theme/base/variables';
+@import '~theme/base/mediaqueries';
+@import '~theme/base/utils';
+
+.footerMain {
+  width: 100%;
+}
+
+.footerInner {
+  padding: 1em 0;
```
emmawedekind: Can we be consistent and use `px` for margin & padding, please? :)

src/components/footer/footer.scss
```diff
@@ -0,0 +1,34 @@
...
+  border-top: 1px solid $color-grey-medium;
+  display: flex;
+  flex-flow: row;
```
emmawedekind: Isn't `flex-direction: row` by default?

emmawedekind:
This looks really great - thanks for your contribution! I just left a couple of minor comments :)

src/components/footer/footer.scss
Venablena: Yes 😁, good catch, it's a leftover from trying to make this mobile-first and then reverting back.

src/components/footer/footer.scss
Venablena: Oups, I think I saw em in someone else's paddings and got it from there. I'll fix those ones too then ;-). (The proofreader that secretly lives in me is excited)

src/components/footer/footer.scss
emmawedekind: No worries! :)

src/components/footer/footer.scss
emmawedekind: I've been trolling everyone else too, don't worry!

src/components/footer/Footer.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { translate } from 'react-i18next';
+import 'i18n/i18n';
+import styles from './footer.scss';
+
+const Footer = ({ t }) => (
+  <footer className={styles.footerMain}>
+    <div className={styles.footerInner}>
```
crysfel: Just wondering if we really need the extra div.... can we use only the `<footer>`?

Venablena: It allows creating margins for that line and keeping the overall content centered while the text is on the right (otherwise the whole line goes to the right or there's no margin). Also, I thought it wouldn't hurt if we decide to add more stuff or a background color to the footer later.

crysfel: that makes sense, thanks!

Venablena:
@emmawedekind Fixed the little issues, ready to merge when review approved.

Issue #120, opened 2018-10-03
Cookies


emmawedekind:
## Description

- [x] Add "Cookies" to the footer (Design 2).  When clicked, this should open a modal (Design 3).

- [x] Add the cookie banner fixed on scroll (Design 1)

## Assets
The text can be found [here](https://docs.google.com/document/d/1Bmo9ZWiHNSjJ5ccsKEfiSTpqtC-KIP91IXQiwnsx5m0/edit?usp=sharing).

### Design 1
![image](https://user-images.githubusercontent.com/7671983/46427359-8f8d3d80-c741-11e8-8b8c-f9bb25533cb8.png)

### Design 2
![image](https://user-images.githubusercontent.com/7671983/46427312-68367080-c741-11e8-95fd-97c1e0d60679.png)

### Design 3
![image](https://user-images.githubusercontent.com/7671983/46428205-a765c100-c743-11e8-8358-9a65f2b428e3.png)

Venablena:
Happy to do this -but is someone already working on the footer/has it been added to a pending pull request, I don't see it in the repo?

Venablena:
@linux-nerd If you're creating the modal, the same should be used for this. I can do the designs 1 and 2.

emmawedekind:
Footer task is here: https://github.com/Coding-Coach/coding-coach/issues/122

Venablena:
What's the expected behavior here? Should the overlay be shown every time the page is loaded or do we want to keep something in local storage, for example, to remember who has visited the page already? I don't know what the rules here are at all.

emmawedekind:
The cookies banner should appear at the bottom of the screen (sticky). The modal only appears if they click the link in the footer.

emmawedekind:
I've posted this on Invision. Please see the links in the Slack channel.

Venablena:
@emmawedekind I didn't see this in InVision but the above image is enough, I think :).
@crysfel Do you want to store it in localStorage without expiration (when the icon has been clicked once on any given device/profile, we won't show this again) or use sessionStorage (it will pop up every time a new window is opened)? I think the latter sounds annoying but just want to make sure. Is it enough to just save something noting that the icon has been clicked (getting the user's IP address would require an external library, I believe)?

crysfel:
I'd say to save in the local storage to show it only once. A boolean value is ok to save, `acceptedCookies=true`.

tureey:
`AreCookiesAccepted`? 😋

Issue #116, opened 2018-10-03
Terms & Conditions


emmawedekind:
## Description

Terms & Conditions MUST be created.  I can create a mock-up but we cannot deliver the landing page otherwise.

## Assets
The document can be found [here](https://docs.google.com/document/d/1_blcaj76xQ6swqhcCTuxnvwp1RjLyXv2j8bT-rB4qbU/edit?usp=sharing).

## Design
![image](https://user-images.githubusercontent.com/7671983/46428158-8ac98900-c743-11e8-8516-ca23449b0937.png)

Venablena:
@linux-nerd Since this uses the same modal as Privacy Policy and Cookies, the same modal should be used for all of them and just change the text.

CoryKelly:
Has this issue been taken yet?

emmawedekind:
I don't think so? I can assign you @CoryKelly

linux-nerd:
@CoryKelly If you have not started working on this, I can complete the same on my PR #117

PR #150, opened 2018-10-19
WIP: Feature: Add Privacy Policy Panel

linux-nerd:
# Details

## Description

- [x] Add Privacy Policy Panel to show the privacy policy
- [x] Remove the mandatory flag from Button component's onClick prop
- [x] Add closeAction prop on Panel component
- [x] Configure Panel.Content component to receive classes props
- [x] Add cookies Policy Panel
- [x] Add Terms and Condition Panel

@crysfel, @tureey Can you suggest about the font family to be used?

I can see two more tasks related to panels - #116 and #120 
I will update the PR to incorporate 3rd point of #120 but for that I need the content.
~~I can see #116 is already assigned to @CoryKelly. I assume @CoryKelly will be working on this.~~
As discussed with @CoryKelly I will incorporate TNC panel (#116) in this PR itself.

## Screenshots
screen shot 2018-10-19 at 11 51 16 am

## Issue
#117


src/components/footer/footer.scss
```diff
@@ -33,3 +33,20 @@
...
   }
 }
+
+.bodyBorder {
+  border: 2px solid $color-grey-medium;
+
+  a {
+    color: inherit;
+    text-decoration: none;
+    border-bottom: 1px solid #000000;
```
emmawedekind: Please use variables.

src/components/footer/footer.scss
```diff
@@ -33,3 +33,20 @@
...
+    transition: background 0.15s ease-out, color 0.15s ease-out;
+    cursor: pointer;
+
+    &:hover{
+      background: #69D5B1;
```
emmawedekind: Use color variables for the following two lines.

src/components/footer/Footer.js
```diff
@@ -1,22 +1,62 @@
...
-      </a>
-    </div>
-  </footer>
-);
+    );
+  };
+
+  const doubleLineBreak = (
+    <React.Fragment>
+      <br /> <br />{' '}
```
AndrewMusgrave: Was the space intentional?

```suggestion
      <br /> <br />
```

src/components/footer/Footer.js
```diff
@@ -1,22 +1,62 @@
...
+      <div className={styles.footerInner}>
+        <a href="#terms" className={styles.footerLink}>
+          {t('footer-terms')}
+        </a>
+        <a href="#cookies" className={styles.footerLink}>
+          {t('footer-cookies')}
+        </a>
+        <Panel
+          trigger={createTrigger('privacy')}
+          modal={true}
```
AndrewMusgrave: There is a shorthand for true props.
```suggestion
          modal
```

src/components/footer/Footer.js
```diff
@@ -1,22 +1,62 @@
...
+          closable={true}
```
AndrewMusgrave:
```suggestion
          closable
```

tureey:
@AndrewMusgrave did going points. Could you update the code with those changes please? @linux-nerd

linux-nerd:
This is strange. My latest commit is not shown here. Have anyone faced this issue before?

linux-nerd:
This is strange. My latest commit is not shown here. Have anyone faced this issue before?

linux-nerd:
@tureey, @crysfel This is really strange. I am not able to see my latest commit here.

linux-nerd:
@tureey, @crysfel This is really strange. I am not able to see my latest commit here.

linux-nerd:
@tureey, @crysfel This is really strange. I am not able to see my latest commit here.

linux-nerd:
@crysfel, @emmawedekind  Can you point me to text required for Cookies popup?

tureey:
As we talked, could it be because the github issues? @linux-nerd  Are they now visible?

tureey:
Related to the font-family, I've always loved `Raleway` or `Lato`. They look *clean* and *smooth*. Both have **light** and **bold weight**. And **they are for high quality** readings. I just love'em :heart: @linux-nerd

linux-nerd:
> As we talked, could it be because the github issues? @linux-nerd Are they now visible?

@tureey I am able to see new commits, but travis ci check is not complete. Not sure if its because of same issue.

Venablena:
There's a link in the issue #120 for the cookie policy text.
https://docs.google.com/document/d/1Bmo9ZWiHNSjJ5ccsKEfiSTpqtC-KIP91IXQiwnsx5m0/edit

linux-nerd:
> https://status.github.com/messages still fucked up @linux-nerd

😆 Everything is working normally now.. phew..

emmawedekind:
Do you have examples for libraries for custom scroll bars? The thing I worry about with a custom scrollbar is performance implications and cross-platform compatibility.

tureey:
I think exactly like @emmawedekind about browser compatibility with the scrollbars but, i would make a modla with: header, body and footer. (The footer for actions if required) and the body scrollable, having or not having custom scroll @linux-nerd

crysfel:
@linux-nerd I personally don't like custom scrollbars, performance is always the concern. I think you should use the default.

src/components/footer/Footer.js
```diff
@@ -1,22 +1,67 @@
...
+import Button from 'components/button/Button';
 import styles from './footer.scss';
+import constant from 'config/constants';
 
-const Footer = ({ t }) => (
-  <footer className={styles.footerMain}>
-    <div className={styles.footerInner}>
-      <a href="#terms" className={styles.footerLink}>
-        {t('footer-terms')}
+const Footer = ({ t }) => {
```
crysfel: To help readability, could you use a class instead? There are several functions inside, I think a class would be easier to read

src/config/constants.js
```diff
@@ -10,4 +10,9 @@ export default {
     INSTA_URL: process.env.REACT_APP_INSTA_URL || 'https://www.instagram.com/codingcoach_io/',
     TWITTER_URL: process.env.REACT_APP_TWITTER_URL || 'https://twitter.com/codingcoach_io',
   },
+  privacyPolicy: {
```
crysfel: What about calling this key `legal`, then we can have *terms & conditions* (If needed) as well as the privacy policy

src/components/footer/Footer.js
```diff
@@ -1,22 +1,67 @@
...
+            i18nKey={detail}
+            lineBreak={doubleLineBreak}
+            freePrivacyPolicy={freePrivacyPolicyLink}
+          />
+        </p>
+      </Panel.Content>
+    </Panel>
+  );
+
+  const doubleLineBreak = (
```
crysfel: Do we really need this? Can we do it with css only?

src/components/footer/Footer.js
```diff
@@ -1,22 +1,67 @@
...
+
+  const closeBtn = <Button>{t('close')}</Button>;
+
+  const freePrivacyPolicyLink = (
+    <a target="_blank" href={constant.privacyPolicy.PRIVACY_POLICY_URL} rel="noopener noreferrer">
+      {t('free-privacy-policy')}
+    </a>
+  );
+
+  const classes = { content: styles.bodyBorder };
```
crysfel: Please move this outside of the component

src/components/footer/Footer.js
```diff
@@ -1,22 +1,67 @@
...
+
+  const footerItems = [
```
crysfel: You can also move this outside of the component, content is not dynamic.

src/components/footer/Footer.js
```diff
@@ -1,22 +1,67 @@
...
-        {t('footer-privacy')}
-      </a>
-    </div>
-  </footer>
-);
+    );
+  };
+
+  const createPanel = (type, detail) => (
+    <Panel trigger={createTrigger(type)} modal closable closeAction={closeBtn} key={type}>
```
crysfel: I think we don't need a `closeAction` prop, we should define a `<Footer><Button>Close</Button></Footer>` element as a child of this `<Panel>`

src/components/panel/Panel.js
```diff
@@ -141,6 +143,11 @@ export default class Panel extends React.Component {
               <section className={mainSectionClasses}>
                 {this.renderCloseIcon()}
                 {children}
+                {closeAction ? (
```
crysfel: Please remove this

src/components/panel/Panel.js
```diff
@@ -48,6 +48,7 @@ export default class Panel extends React.Component {
     open: bool,
     modal: bool,
     classes: object,
+    closeAction: node,
```
crysfel: Please remove this prop

src/components/panel/PanelContent.js
```diff
@@ -1,8 +1,28 @@
...
 import styles from './assets/panelcontent.scss';
 
-const Content = ({ children }) => {
-  return <section className={styles.mainContent}>{children}</section>;
+const cx = classNames.bind(styles);
+
+const Content = ({ children, classes }) => {
+  const contentClasses = cx({
+    mainContent: true,
+    [classes.content]: classes.content,
```
crysfel: Could you rename this to `classes.root`? Just to be consistent with the rest of the components.

src/i18n/en.js
```diff
@@ -26,6 +26,13 @@ export default function loadEn() {
       'contact-page-media-alt':
         'Person using a magnifying glass to see the details of another guy.',
       'slack-org': 'Slack Organization',
+      close: 'Close',
+      'privacy-policy': `Effective date: October 03, 2018 {{lineBreak}}
```
crysfel: Do we really need `{{lineBreak}}` here? Maybe we can just do it with css?

src/components/button/Button.js
```diff
@@ -31,7 +31,7 @@ Button.propTypes = {
   children: PropTypes.node.isRequired,
   disabled: PropTypes.bool,
   fullWidth: PropTypes.bool,
-  onClick: PropTypes.func.isRequired,
+  onClick: PropTypes.func,
```
crysfel: Do we need to make this optional? I feel like having a button that doesn't do anything is not a good UX xD

src/components/footer/Footer.js
linux-nerd: Yes we can do that, but if we do that, then we need to maintain state of the panel -> open/close to toggle the panel. And for that we need to make it a container component. Right now in the footer we need to show 3 panels. If we remove this prop then we need to maintain 3 states of the panel and change the presentational footer component to presentational component.

Moreover, I feel this is really beneficial if no specific action is to be taken on closing the panel. If there is a need to perform any action then we can use `<Footer><Button onClick={abc}>Close</Button></Footer>`

Let me know your thoughts

src/components/footer/Footer.js
linux-nerd: I am not sure. I will be more than happy if we can 🤞

src/components/button/Button.js
linux-nerd: I don't think we should enforce onClick prop on button component. 
There might be a case where I just want to pass the button as a trigger node to the panel without any onClick handler defined. And later on the Panel component adds its onClick functionality to the Button component.

linux-nerd:
@emmawedekind can you explan cross

linux-nerd:
I really don't know how I closed it. 😮

linux-nerd:
Reopening this accidentally closed PR

linux-nerd:
> Do you have examples for libraries for custom scroll bars? The thing I worry about with a custom scrollbar is performance implications and cross-platform compatibility.

I just came across one - http://malte-wessel.github.io/react-custom-scrollbars/

In my previous projects I have used perfect-scrollbar (angular 1.x and 2+) and I never encountered any performance issue on desktop/mobile/tablets.

The driving force for custom scrollbar is same look and feel across different browsers and OSs

src/components/footer/Footer.js
linux-nerd: Should we have rich text for policies (cookies, TnC etc.) instead of plain text. Its really a pain to style plain text 😞 
But I am not sure if this is correct approach (due to XSS attack) Any thoughts @crysfel @tureey @emmawedekind

src/components/footer/Footer.js
linux-nerd: I don't really like using classes for presentational component, may be I can move all the functions out of the component, and since its ES6 modules the variables will not leak out of the module. What say?

crysfel: Yes, please do that.

tureey: why not use a simple `function Footer({ t })`?

Nothing is going to change inside, so it's better than a `function expression` isn't it? In this case, I prefer: `function` > `stateless` > `statefull`

src/components/panel/Panel.js
tureey: I did not see this! What is used for? The name makes sense for a function but for a node? I am curious

src/components/panel/PanelContent.js
crysfel: Right, but I was thinking to always have a `root` class for the root element on every component, then inner elements can have any other name.

The backdrop of the panel is in a different component, right?

src/i18n/en.js
tureey: No idea either

crysfel: For titles is fine, I was just thinking about the line breaks.

src/components/button/Button.js
crysfel: that sounds legit 👌

tureey: I can't imagine the scenario by now. I would say to make it a mandatory parameter. If one day happens what you say, we will talk about this :thinking:

crysfel:
@linux-nerd maybe we can add it later, we can consider it. But for now I think we are good to go

src/components/panel/Panel.js
linux-nerd: @tureey I will try to explain :)
If we want to add a close button to the panel - 
```
<Panel
    trigger={<Button text="Open Modal" />}
     closable={true}
     modal={true}
     open={this.state.open}>
        <Panel.Header>Hello World!</Panel.Header>
        <Panel.Content>
            <p>To close the modal, use any of the buttons/actions</p>
         </Panel.Content>
         <Panel.Footer>
            <Button text="Deny" onClick={this.closeModalHandler} />
            <Button text="Accept" onClick={this.closeModalHandler} />
          </Panel.Footer>
</Panel>
```

Let's concentrate on `Panel.Footer` component.
It has two button components which handle the close action. The purpose of these buttons is to only close the panel and nothing else. For this we need to maintain the state of the panel in our component.

Consider a situation when we have multiple panels in the same component. We will have to change our component to stateful component (`class Comp extends React.Component{}`) and manage all the states in our Comp component. I feel this is an overhead to manage the state of Panel in the component, if we do not intend to do any action on the close of the panel.

I can change the name to something else.

Let me know if you have any questions.

src/components/button/Button.js
linux-nerd: @tureey 
Let's take a simple example - 
```
<Panel
    trigger={<Button text="Open Modal" />}
     closable={true}
     modal={true}
     open={this.state.open}>something</Panel>
```

Over here the `trigger` prop is a button. This button will have a onClick prop in the panel component, but not here. If Button has onClick as mandatory, then we will have to unnecessarily change it to `<Button onClick={noop} text='Open' />` which is a hack to overcome the limitation of Button component.

src/components/footer/Footer.js
linux-nerd: @tureey What is the difference between `function` and `stateless` components?

linux-nerd:
> @linux-nerd maybe we can add it later, we can consider it. But for now I think we are good to go

Regarding custom scrollbars?

emmawedekind:
What's the status on this? Can we get this finished up and merged so we can release soon? :)

src/i18n/en.js
```diff
@@ -26,6 +26,25 @@ export default function loadEn() {
         'We believe that mentorship should be accessible to all people regardless of location or financial standing. In pursuit of this goal we will provide a free and open source platform to facilitate mentorship connections. Our mission is to foster a greater sense of collaboration and inclusiveness in the technical industry by providing a platform to aid the mentorship process.',
       'home-mission-media-alt': 'Group of people',
       'slack-org': 'Slack Organization',
+      close: 'Close',
+      'privacy-policy': `Effective date: October 03, 2018 {{lineBreak}}
```
linux-nerd: @crysfel Do you like this approach ?
I feel it's little ugly 😞

src/i18n/en.js
```diff
@@ -26,6 +26,25 @@ export default function loadEn() {
...
+      </1>
+      <2/>
+      As is common practice with almost all professional websites this site uses cookies, which are tiny files that are downloaded to your computer, to improve your experience. This page describes what information they gather, how we use it and why we sometimes need to store these cookies. We will also share how you can prevent these cookies from being stored however this may downgrade or 'break' certain elements of the sites functionality.
+      <3/>
+      <4/>
+      For more general information on cookies see the Wikipedia article on HTTP Cookies.
+      <5/>
+      <6/>
+      <7>How We Use Cookies</7>
+      `,
```
linux-nerd: @crysfel Do you like this approach ?
I feel it's little ugly 😞

src/i18n/en.js
linux-nerd: screen shot 2018-10-25 at 7 21 24 pm

@crysfel I don't think css option is working very well.

src/components/panel/Panel.js
tureey: I think we mixed the idea of a Modal with a Panel. **For me, these are 2 different concepts.** A modal is something that can be **shown** or **hidden** and Panel is always visible.

Modal should be no more than one. Why? Because it's a portal. It's no more than a `<div>` that is used as a widget to inject HTML on demand. *In my opinion, we should have a HOC with functions that opens the portal given a title, a content and actions. and this hoc should be able to render the content provided to the Portal.*

To be pragmatic, I would talk about this once we have the problem you said and release with the changes that @crysfel was saying. We should **remember that iteration is the key**.

src/i18n/en.js
crysfel: Given that the content has formatting and styles, I was thinking that maybe instead of putting all that in a the i18n files, maybe just creating `markdown` files that we can load dynamically and parse when needed.

After all people is not going to open this very often and it will increase our bundle size and is adding some unnecessary complexity to our codebase. 

Anyway, we can iterate on this. Let's create a separate issue for this and merge the PR as it is, it doesn't look great but we need to deliver something soon :)

src/i18n/en.js
crysfel: Let's just go with the linebrakes and everything inside the i18n files, then we can iterate and come up with a better solutions, that maybe includes loading the content async, after all users won't open this modal very often

src/components/panel/Panel.js
crysfel: > To be pragmatic, I would talk about this once we have the problem you said and release with the changes that @crysfel was saying. We should remember that iteration is the key.

👌

src/i18n/en.js
tureey: Yes, I agree with the markdown. It's something I commented below. It's going to be easier for future static pages. I would leave this how it's now and then.. have something like `markdown->JSX` converter.

src/components/button/Button.js
crysfel: My personal preference is to make it required as it was before.

crysfel:
> > @linux-nerd maybe we can add it later, we can consider it. But for now I think we are good to go
> 
> Regarding custom scrollbars?

Yes, regarding the scrollbars

linux-nerd:
@crysfel, @tureey, @emmawedekind I am stuck with some urgent work for today and tomorrow. I can resume my work on this only on Sunday.

If anyone of you have some time, feel free to proceed with it.
Pending Items - 

- [ ] styles are broken for mobile devices
- [ ] Content for TnC and Cookies policy
- [ ] Some code review comments. I think I am not able to explain my point for adding closeAction prop in the panel and removal of required attribute from Button. May be we can collaborate for that.

src/components/button/Button.js
tureey: I do agree with you @crysfel

tureey:
I am going to try to fix the styles now :smile:

tureey:
Developed here https://github.com/Coding-Coach/coding-coach/pull/175

crysfel:
#175 continues the work on this PR, therefore I'm closing this

PR #162, opened 2018-10-20
Create consistent folder structure

AndrewMusgrave:
# Details
A consistent file architecture.

## Description
A purposed consistent file structure. `src/components` will have an index file containing all of the component exports. Components can now be imported as 
```jsx
import {A, B} from 'components';
```
 rather than
```jsx
import A from 'components/A/A';
import B from 'components/B/B';
```
Components have a `__tests__`, `assets/images`, `stories` and `component` folder. The component folder will hold the sub components with their respective components, tests, etc.

## Steps to QA

- [x] Step 1 - Do the tests pass?
- [x] Step 2 - No changes in storybook
- [x] Step 3 - Visually check development server
- [x] Step 4 - Run `yarn build`

## Screenshots

filestructure

## Issue

There was no issue. This adds a consistent, easy to follow file structure.


AndrewMusgrave:
Looks like the tests are running with the incorrect paths. For example
```js
src/components/panel/components/Footer/_tests_/PanelFooter.test.js
```
should be
```js
src/components/Panel/components/Footer/_tests_/PanelFooter.test.js
```

I have a suspection this has something to do with the failing tests, everything is 👌 locally. I'm not familiar with travis-ci, any ideas?

crysfel:
Thanks for helping with the consistency of our codebase.

I'm personally not in favor of importing components directly from `components`, instead I prefer:

```
import A from 'components/A';
import B from 'components/B';
```
A couple reasons:

- When removing a component (because of cleanup), it's fairly easy to find out where the component is being imported, a simple search will do the trick. Where if we use `import {A, B} from 'components';`, it is more challenging to find all the imports.
- Same issue when updating the name of the component

tureey:
Well, I would only export the Component as you say, but creating the `index.js` and not naming the file as the component. So it's cleaner and faster to import.

As you point, I prefer to export only the component but in case it's a namespace, we should export all the components (like the panel for example). Am I right? Or is there another way to do it?

crysfel:
> Well, I would only export the Component as you say, but creating the `index.js` and not naming the file as the component. So it's cleaner and faster to import.

Yeah, this is how I like and I usually do it. But we decided to go with the current file name because not everyone is using VSCode and opening many `index.js` files at the same file is confusing 🤷‍♂️

> As you point, I prefer to export only the component but in case it's a namespace, we should export all the components (like the panel for example). Am I right? Or is there another way to do it?

What I like about this PR is the addition of the `index.js` for every component, and that is exporting the original component, it will help with the imports indeed. 

I feel like is redundant to have `index.js` and `ComponentName.js`, however... this is the only way every one will be happy, because once the `index.js` is created, there's no need to touch it again.

My preference:
+1 to have an `components/component/index.js` for every component!!
-1 to have `components/index.js` exporting all components xD

I'd like to hear other people opinions! I'm always open to new ideas, if everyone is ok to have `components/index.js` exporting all component, I'm ok! But I want to hear how everyone feel about it!

tureey:
*What I like about this PR is the addition of the index.js for every component, and that is exporting the original component, it will help with the imports indeed.*

Can't reply using the phone 😂. Does this mean that even components with only 1 component wikk have the `index.js` and the `Component.js`?

linux-nerd:
> > What I like about this PR is the addition of the index.js for every component, and that is exporting the original component, it will help with the imports indeed.
> 
> Does this mean that even components with only 1 component will have the `index.js` and the `Component.js`?

Exactly, I was about to ask the same thing..

I have seen many libraries using the barrel for exposing the components/functionalities. In my opinion I don't feel the need of introducing the concept of barrel/index.js as of now. If we talk about Panel, the header/footer/content can be accessed via  Panel.Header/Panel.Footer/Panel.Content as they are static properties for Panel Component and I don't see them being used individually. 

If we go ahead and introduce barrel to Panel or any other component then we should introduce it for rest of the components as well. As @crysfel said I am also in favour of not introducing barrel/index.js at component level.

linux-nerd:
I am also not in favour of adding another level of hierarchy of components -> `components/panel/components/Footer/Footer.js`. I feel this is really high level of abstraction. The current structure for components is just fine, except that we need to introduce a strict rule for placing the css/scss files.

I see two types of followers for placing the stylessheets ->

1.  Next to component
2. Nest it inside assets/

I like the placement of stylesheets inside assets directory per component.

tureey:
I don't. We will have tons of folders of assets only because of the scss..

mrassili:
I prefer that you have each set of components in a directory, and that should contain all related components (as well as their test & css files). Plus when importing, you know what each component is related to, instead of just 'components'

PR #180, opened 2018-10-27
tests: test Node.js 8, 10 and 11 (stable)

DanielRuf:
# Details

## Description

Also test Node.js 8, 10 and 11 (stable).


.travis.yml
```diff
@@ -1,7 +1,9 @@
 dist: trusty
 language: node_js
 node_js:
-  - "10.10.0"
+  - "8"
```
crysfel: The app doesn't run on node 8.... I wonder why Travis is not complaining

DanielRuf: Is this made clear in package.json? I see nothing regarding `engines`. Also it looks like all dependencies work under 8 and newer.

tureey: It should be mandatory to run in node 10.x :thinking:

DanielRuf: > It should be mandatory to run in node 10.x 🤔

Any good reason why?

DanielRuf: Please list the packages which do not work on Node.js 8 which is still LTS (like Node.js 6) and supported until 2019.

tureey: I don't know why lower versions than`10.x` can't transpile correctly and they throw errors when not coding as conventional way.

For example, state without constructor.

(wait for @crysfel )

DanielRuf: > they throw errors

Would be good to know which errors exactly so I can check the ES support tables of Node.js against this.

tureey: I'll look for them on the closed PR's and paste them here.

crysfel: Node 8 doesn't support class properties, like these:

```
class SomeComponent extends Component {
  state = {
    test: 'testing',
  };

  onClick = () => {

  }
}
```

> Is this made clear in package.json? I see nothing regarding engines.

Can you add it in this PR?

Thanks!

DanielRuf: > Node 8 doesn't support class properties, like these:

Would be new to me as I've used classes like this in past React projects.

I'll check this and probably add the constraint to the `engines` field.

DanielRuf: See https://node.green

tureey: First time I see this page. lol **Thank you**

crysfel: > Would be new to me as I've used classes like this in past React projects

Interesting... we were having errors while building with Node 8, migrating to 10.x solved the problems.

linux-nerd: > Node 8 doesn't support class properties, like these:
> 
> ```
> class SomeComponent extends Component {
>   state = {
>     test: 'testing',
>   };
> 
>   onClick = () => {
> 
>   }
> }
> ```
> > Is this made clear in package.json? I see nothing regarding engines.
> 
> Can you add it in this PR?
> 
> Thanks!

@crysfel How will this affect the build of the application. I think it will transpile correctly. Did I miss something?
I am using node v9.9 and not 10. The only issue I have is with yarn install. I have to pass --ignore-engines for yarn to install the dependencies correctly.

tureey: If im not wrong, Travis can't build up the application. It means that the PR is going to be blocked @linux-nerd I think

DanielRuf: Not sure what you mean with blocked. The tests / CI builds are green.

DanielRuf: Not sure what you mean. In this case you just test the supported engines (Node.js 10 and 11).

DanielRuf: Node.js 9 should be already EOL.

crysfel: I agree, but for some reason this was not the case. This is the first time I'm using CRA, usually I like to have full control over my build process, seems like CRA requires node `8.10.0` and above I will test later today with that version and see if everything works fine.

reeversedev: @crysfel @DanielRuf I don't know if this is relevant today or not but while searching for an issue regarding core-js modules, I came to know a fact that Node.js 10.x doesn't support 32-bit architecture (i386) which can actually affect many developers out there!
That being said, there can be problems on different systems for developers.

DanielRuf: @reeversedev do you have any reference and links about this ussue? Never had or saw the described issue (why uses a 32 bit system today?).

DanielRuf: But in general, how is this relevant for the CI tests which do not affect users and contributors?

emmawedekind:
What's going on here?

PR #185, opened 2018-10-30
Feature/refactor buttons

tureey:
# Details

Due to the code in button could be unified and DRY, that's my proposal. (for sure it's optimizable, so please, don't hesitate on improving this!). Thank you 😄 

## Description

Trying to DRY code a little bit.


src/components/button/button.scss
```diff
@@ -1,6 +1,31 @@
...
 
+
+@mixin btn-box-shadow ($color) {
+  box-shadow: 0px 7px 10px rgba($color, 0.5);
+
+  &:hover, &:focus {
+    box-shadow: 0px 4px 5px rgba($color, 0.6);
+  }
+
+  &:Active {
```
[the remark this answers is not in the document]
tureey: uops, typo! Good catch

src/components/button/Button.js
```diff
@@ -21,7 +21,7 @@ const Button = (props) => {
       className={classNames(
         styles.button,
         styles[type] || styles.primary,
-        styles[size] || styles.default,
+        styles[size] || styles.mmedium,
```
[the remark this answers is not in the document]
tureey: yes! good catch again!

src/components/button/button.scss
```diff
@@ -35,72 +58,67 @@
...
-  // For icons and svgs
   i, svg {
     vertical-align: middle;
     margin-right: 4px;
   }
 }
 
-.default {
+.small {
+  composes: PHS;
```
nitish24p: sorry but what's PHS

src/components/button/button.scss
```diff
@@ -1,6 +1,31 @@
...
+  &:hover, &:focus {
+    box-shadow: 0px 4px 5px rgba($color, 0.6);
+  }
+
+  &:Active {
+    box-shadow: none;
+  }
+}
+
+@mixin btn-borders ($color, $color-darker) {
```
nitish24p: 👍

nitish24p:
Can you add a rounded prop, or a type of pill or box or some prop to handle rounded button too.. i was wondering if you're refactoring button maybe you could do this too iff possible

src/components/button/button.scss
tureey: @crysfel created a way to include paddings and margins.
For mobile first styles:
```
// This adds padding horizontal small
// padding-left: small
// padding-right: small

composes: PHS;
```

**IMPORTANT:**
Inside mediaqueries it does not works. For these cases, we have a mixin called `spacing`
It's all in the file called `_spacing.scss`

tureey:
Related to this prop you are talking about @nitish24p , I proposed a prop called `shape` which can have `square`, `rounded`, `pill`

so the definition would be:
```
Button.propTypes = {
    shape: PropTypes.oneOf(['square', 'rounded', 'pill']),
}
```

Maybe there is some people that say: `square` will make us thing that maybe it's a square button, so it has the same width and height. In this case we could call the prop `rectangular`

@crysfel @emmawedekind

src/components/button/button.scss
crysfel: PHS stands for Padding Horizontal Small, which translate to:

```
.PHS {
  padding-left: 8px;
  padding-right: 8px;
}
```

It's basically a way to keep all our spacing consistent across the app.

crysfel:
@tureey I think a `shape` prop is self explanatory and give us the flexibility to add more shapes later on. The tradeoff I see is that it might be confusing having many different shapes, but we can mitigate that with the stories showing all the values.

I'd would say, go ahead and implement it!

tureey:
Tomorrow I'll do it then and update the PR

src/components/button/Button.js
```diff
@@ -21,7 +22,8 @@ const Button = (props) => {
       className={classNames(
         styles.button,
         styles[type] || styles.primary,
-        styles[size] || styles.default,
+        styles[size] || styles.medium,
```
crysfel: I don't think the `or` is necessary here, we have a default value and we have a validation to only accept certain values. Therefore `size` will always have a valid value. 

Please remove all these `||`

src/components/button/stories/button.README.md
```diff
@@ -48,5 +49,21 @@
 </Button>
 ```
 
+#### Button Shapes
+```jsx
```
crysfel: Thanks for documenting this, it'd be nice to also have a working demo in the `Button.stories.js` file, can you add them there as well?

src/components/button/Button.js
tureey: I thought the same when I added the ones related to `shape`. Actually I wanted to remove them, but I was not sure... I'll delete it tomorrow!

src/components/button/stories/button.README.md
tureey: sure! Tomorow i'll do it

tureey:
@crysfel check again and merge

Issue #188, opened 2018-10-30
UI shifts when a modal is opened


emmawedekind:
Re-creation steps:
1. Click Terms & Conditions link in footer

Expected behavior: UI in the background should remain in the same position.
Actual behavior: UI in the background shifts a couple of pixels to the right.

![image](https://user-images.githubusercontent.com/7671983/47732533-ac486280-dc66-11e8-9df6-f73760fdd698.png)

![image](https://user-images.githubusercontent.com/7671983/47732542-b1a5ad00-dc66-11e8-8c28-f460f51e3af1.png)

tureey:
Yes, that's actually made on purpose by me... I wanted to make sure we had no problems with the body scroll. It's annoying when you scroll and it affects to the body content. that's why..

To evade this, I added a class to the body called `u-overflowHidden`. this class makes:
```
.u-overflowHidden {
    height: 100vh;
    overflow: hidden;
  }
```

But if you guys think its better to do not have this "movement" on the window, just remove it ^^

linux-nerd:
Updated and PR - #196 raised.

PR #196, opened 2018-10-31
Fix: #188 UI shifts when a modal is opened

linux-nerd:
# Details

## Description
Removed the style overflow hidden when the modal is open. I have tested it only on mac (Chrome, FF and safari). Need someone to confirm on windows as well.

## Issue
#188


emmawedekind:
Closes #188

Issue #195, opened 2018-10-31
The user is taken to different sections without animation.


linux-nerd:
# Issue

## Description

When Learn More is clicked, it animates and take the user to About Section. But when the user navigates from the header links -> About/Mission/Contact. The user is taken to these sections without animation.

The behaviour should be consistent.

## Direct Requirements

- on click of the menu items, the user should be taken to the desired section with animation.

tureey:
The call the action (button) takes you smoothly to the about sextion because it uses a scrollIntoView. The click event is catched and then we scroll.

Same logic would have to be implemented.

tureey:
Shit sorry hahahah

linux-nerd:
> Shit sorry hahahah

😆

linux-nerd:
PR - #201 raised

crysfel:
Thanks!

PR #201, opened 2018-10-31
FIX: #195 The user is taken to different sections without animation

linux-nerd:
# Details

## Description

- [x] Added scrollIntoView to smoothen the view
- [x] Added id to the url hash

## Issue
#195

### DIFF
--- a/src/components/navbar/Navbar.js
+++ b/src/components/navbar/Navbar.js
@@ -3,16 +3,26 @@ import { translate } from 'react-i18next';
 import 'i18n/i18n';
 import styles from './navbar.scss';
 
+const handleClickOnNavigation = (id) => {
+  return (event) => {
+    event.preventDefault();
+    document.getElementById(id).scrollIntoView({
+      behavior: 'smooth',
+    });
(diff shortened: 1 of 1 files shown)


src/components/navbar/Navbar.js
```diff
@@ -3,16 +3,26 @@ import { translate } from 'react-i18next';
 import 'i18n/i18n';
 import styles from './navbar.scss';
 
+const handleClickOnNavigation = (id) => {
+  return (event) => {
+    event.preventDefault();
+    document.getElementById(id).scrollIntoView({
+      behavior: 'smooth',
+    });
+    window.history.pushState({}, '', `/#${id}`);
```
tureey: There is an issue created to don't upadte the url so we should use buttons instead of anchors. Maybe we could try to do both in this PR

crysfel: I think that issue is only for the modal window links. I think is ok to set the hashes when going to these sections.

linux-nerd: @tureey I created that issue for modals. I added the id to the url to make the bookmarking possible.

PR #208, opened 2018-11-02
Fixed navbar at narrowest screen widths

### LINKED ISSUE #204: Navbar spills over onto second line on narrowest screens
# Issue

## Description

Navbar spills over onto second line on narrowest screens, this was first noticed using an iPhone SE and after investigating this issue seems to occur almost exactly at the 320px breakpoint.

## Screenshot

alt text

## Direct Requirements

- [x] Add media query to handle display of Navbar on narrowest screens

msweeneydev:
# Details

## Description

Previously, the navigation bar (due to its inner padding) would span two lines at the smallest mobile resolutions, below 320px (iPhone SE etc.).

This fix adds a media query to address screens below this size and targets those devices, using flexbox to space the items as well as overriding the gaps between them.

With this in place, we can handle gracefully a 3 item navbar all the way down to a 250px width device, I am not aware of any devices smaller than this.

## Screenshots

alt text

## Issue

Closes #204

### DIFF
--- a/src/theme/base/_mediaqueries.scss
+++ b/src/theme/base/_mediaqueries.scss
@@ -32,17 +32,11 @@
   }
 */
 
-$mobile-and-smallest: 320px; // iPhone SE & smaller
 $tablet-and-wider: 768px; // Tablet
 $desktop-and-wider: 1024px; // Desktop
 $desktop-large-and-wider: 1440px; // Large desktops
 
 @mixin respond-to($media) {
-  @if $media == mobile {
-    @media screen and (max-width: $mobile-and-smallest) {
-      @content;
-    }
-  }
   @if $media == tablet {
     @media screen and (min-width: $tablet-and-wider) {
         @content;
--- a/src/theme/base/_mediaqueries.scss
+++ b/src/theme/base/_mediaqueries.scss
@@ -32,11 +32,17 @@
   }
 */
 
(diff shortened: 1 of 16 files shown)


src/theme/base/_mediaqueries.scss
```diff
@@ -32,11 +32,17 @@
   }
 */
 
+$mobile-and-smallest: 320px; // iPhone SE & smaller
 $tablet-and-wider: 768px; // Tablet
 $desktop-and-wider: 1024px; // Desktop
 $desktop-large-and-wider: 1440px; // Large desktops
 
 @mixin respond-to($media) {
+  @if $media == mobile {
```
crysfel: This might be confusing, we are building the app following a mobile first approach, adding a `mobile` media query might confusing I think.

What if we do this for all mobiles:
```
    display: flex;
    justify-content: space-evenly;
    padding: 0 2rem;
```
I think it will look OK on larger mobile screens.

tureey: I think the same

crysfel: yes, I think that would be better.

msweeneydev:
Updated as discussed, it now reverts to how it was pre-change at the 768px breakpoint using the existing tablet query.

PR #211, opened 2018-11-05
Fix/footer links

### LINKED ISSUE #197: Modal when opened changes the url
# Issue

## Description

Whenever a modal window is opened, the url gets appended with `#something`. In case of popup this dosen't make sense.

## Screenshot

screen shot 2018-10-31 at 9 26 35 am

screen shot 2018-10-31 at 9 26 06 pm

## Steps to Reproduce

1. Click on Cookies Policy
2. Observe the url

## Direct Requirements

- Do not update the url
- Please apply same fix to Privacy Policy, Terms and Condition and Cookie Policy in the footer

nero2009:
# Details
Changed footer links from a tag links to Button component

## Description
Changed footer links to button component because when the modal opens the url is appended to it. Also the default text for the Button component is in Uppercase so I added a new `prop` called `typography` to the Button component with can be used to specific the case of the text. Finally I removed the  test for the `Footer` component because it expects the footer links to be anchor tags.

## Issue

closes #197

### DIFF
--- a/src/components/footer/footer.scss
+++ b/src/components/footer/footer.scss
@@ -41,6 +41,7 @@
   font-size: $font-xxs;
   color: $color-grey-medium;
   text-decoration: none;
+  text-transform: none !important;
   &:hover {
     color:$color-grey-dark;
   }
--- a/src/components/footer/footer.scss
+++ b/src/components/footer/footer.scss
@@ -41,7 +41,6 @@
   font-size: $font-xxs;
   color: $color-grey-medium;
   text-decoration: none;
-  text-transform: none !important;
   &:hover {
     color:$color-grey-dark;
   }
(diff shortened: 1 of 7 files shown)


src/components/footer/footer.scss
```diff
@@ -41,6 +41,7 @@
   font-size: $font-xxs;
   color: $color-grey-medium;
   text-decoration: none;
+  text-transform: none !important;
```
crysfel: This looks interesting.... the `Button` component is defining `typography="capitalize"` and then here is removing that? Can we create a `none` option to the `typography` prop that doesn't set any transformation to the text?

The thing is that using `important` statements in css tends to lead to css issues later down the road.

nero2009: Ok I will do that

crysfel:
Woops.... there are some conflicts, will merge after resolving the conflicts, thanks!

PR #249, opened 2018-12-01
Feature/migrate modal styles to tail wind

### LINKED ISSUE #236: Migrate Modal component to tailwind
# Issue

## ZenHub Issue
The current Modal component is using SASS for styling, we need to remove all that code and use tailwind instead

tureey:
# Details
There are a few styles I don't know how to migrate:
`background-color` with opacity, `max-width`, `max-height`, `min-width` with values different than `0` or `100%`

## Description

Migrate the styles of the `Modal Component`

## Issue

https://github.com/Coding-Coach/coding-coach/issues/236

Closes #236


src/components/modal/modal.module.scss
```diff
@@ -1,27 +1,11 @@
...
-    top: 0;
-    left: 0;
-    width:100%;
-    height: 100%;
-    background: rgba($color-black, .3);
-    display: flex;
-    align-items: center;
-    justify-content: center;
-}
+.bgColor { background: rgba($color-black, .3); }
```
crysfel: We might want to create a plugin for this, to allow using this same styles somewhere else.

tureey: Yes, I agree!

- Does it mean the rest of the PR is okay? *if that's the case,*
  - Does this line that you commented should block the PR?
  - Should we merge and create tech debt?

crysfel: Yeah... I'm thinking in the plugin we should move all the styles related to the modal, and use the tailwind utilities there to set the responsive styles.

tureey: I'll be waiting for news then! Okay? @crysfel

tureey:
So, how we operate with the "plugin" thing? @crysfel

tureey:
@crysfel I prefer to do in a new one

tureey:
I added the modal background color in the palette. Maybe it's not the best practice. We'll have to add a component plugin maybe? @crysfel Let's close the branch if the rest of the things are OK and create another task to do it if that's okay

Issue #258, opened 2018-12-09
Components - Notifications


emmawedekind:
![image](https://user-images.githubusercontent.com/7671983/49695125-51bff180-fb96-11e8-9fa9-4648fc25d0b0.png)

- [ ] Mentorship notifications have blue edge

- [ ] Account notifications have green edge

- [ ] Error and issue notifications have red edge

- [ ] Selected notification has clear button appear in top right corner and a box shadow effect

# Assets
All icons & color codes can be found on the Sketch cloud: https://sketch.cloud/s/xVR5A/p/notifications

emmawedekind:
Sure, I'll assign you! Looks like you weren't added to the organization, so I went ahead and added you. I'll assign you to the issue once your addition has been registered. Thanks so much!

alexluong:
@emmawedekind Thanks for the invitation. I joined the organization.

crysfel:
@alexluong all yours! You should use storybook for this, as we don't have the dashboard ready.

alexluong:
@crysfel Thanks for the tips. I've been using storybook for it. Can you check out my ongoing PR #261? I got a few questions over there!

alexluong:
Let's close this issue. I'd love to help out more if there's any other tasks!

crysfel:
@alexluong thanks for contributing!

PR #261, opened 2018-12-09
Notification component

alexluong:
#258


alexluong:
Hi folks,

I'm working on the Notification component. This is still on-going. I'd like discuss about how we want to design the API of the component a bit to make sure that it works well with our current app

Before starting on this ticket, I made a PR #260 to set up Storybook. **Therefore, we may have to merge that ticket first.**

First of all, I add a new color to `tailwind.js`, `blue: '#1f386e'` for the mentorship color.

--- **Do we want to name it differently or set up a whole blue variations of blues?**

---

Now, for the `Notification` component. Here is the current props it's expecting:

```
| propName | propType | defaultValue | isRequired | Description                |
| -------- | -------- | ------------ | ---------- | -------------------------- |
| onClick  | function |              | no         | Click Handler              |
| title    | string   |              | yes        | Notification title         |
| message  | string   |              | yes        | Notification message       |
| time     | string   |              | yes        | Notification time          |
| type     | string   | primary      | no         | mentorship, account, error |
```

And this is how it's looking:

```jsx
<Notification
  type="mentorship"
  title="Emma wants to be your mentee!"
  message="Head over to invitations to check out your invite."
  time="10:29 p.m."
/>
```

I have a few questions regarding this:

--- How do we want to pass in the `time`? A configured string like that (`"10:29 p.m."`); a`Date` object; an `Int` Unix time number?

--- Does the component know how to discard the notification? I see that on hover, users can close the notification. Should there be a props `onClose` or something similar?

--- Besides those questions, how is everything looking? Any feedback would be appreciated.

emmawedekind:
This is awesome! @crysfel if you could take a look at the questions whenever you're free, we'd very much appreciate it! :) Have a nice weekend!

emmawedekind:
Just curious... why are we using Font Awesome? I have a subscription to FlatIcon and would prefer to use those SVGs if possible.

alexluong:
Umm it was just easier for me at the time. I couldn't find a Close icon in our code base and were not familiar with Sketch Cloud enough to get the asset out. Can you set up the svg icon, @emmawedekind? That would be really helpful for me. Thanks.

emmawedekind:
I just sent it on Slack :)

crysfel:
@emmawedekind fortawesome was introduced a while ago, there's an issue to remove it and use our custom icons.

alexluong:
@crysfel I switched to our custom icon already. I think this PR is ready to be reviewed and merge.

I also added `date-fns`. I read through the tech stack document, and it did say that it was our library to handle date. Just wanna let you know about it.

src/components/notification/Notification.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
```
crysfel: Just for consistency with the rest of the code base, let's import the prop types like this:

```
import { oneOf, string, ..... } from 'prop-types';
```

src/components/notification/Notification.js
```diff
@@ -0,0 +1,120 @@
...
+  type: PropTypes.oneOf(Object.values(NOTIFICATION_TYPES)),
+  title: PropTypes.string.isRequired,
+  message: PropTypes.string.isRequired,
+  time: PropTypes.instanceOf(Date).isRequired,
+  onClose: PropTypes.func.isRequired,
+  onClick: PropTypes.func,
+};
+
+Notification.defaultProps = {
+  type: 'primary',
```
crysfel: I think `primary` is missing in `NOTIFICATION_TYPES`

tailwind.js
```diff
@@ -48,6 +48,7 @@ let colors = {
 
   'black': '#202939',
   'white': '#ffffff',
+  'blue': '#1f386e',
```
crysfel: Please use a different name for this color, maybe `neutral`, later on we might want to add some variations.

alexluong:
@crysfel I made the changes. Let me know if that works.

PR #280, opened 2019-01-04
ci: test pnpm

DanielRuf:
# Details

## Description

There are some issues with pnpm. The issues with the dependencies can be solved using an npmfile.js. The rest is because of antipatterns in the code.

## Steps to QA

Run `pnpm install && pnpm run lint && pnpm run i18n-extract && pnpm run i18n-compile && pnpm run test && pnpm run build`

## Issue

pnpm found several issues.


DanielRuf:
Also the cp command is not part of the scripts (running on CI and so on) and is not very verbose as these are env vars instead of being part of .npmrc which is used by Yarb too.

DanielRuf:
https://www.npmjs.com/package/config too and so on. Same for the Sass imports.

trent-boyd:
Hmm. Overall, I'm still not really sure. I've only had issues on projects where we allow devs to use two different package managers in a project. I know that's vague, but minor inconsistencies can lead to dev churn time making special accommodations, when it's just easier for all agree to use a single tool and call it a day.

> Also the cp command is not part of the scripts (running on CI and so on) and is not very verbose as these are env vars instead of being part of .npmrc which is used by Yarn too.

Moving the `.env.example` to `.env` is just something mentioned very early in the Readme as part of setup. We could also just commit the `.env` file directly to skip that setup step, but that's not my call to make. Having the `NODE_PATH` in an `.env` file is part of the `create-react-app` config [as documented here](https://facebook.github.io/create-react-app/docs/advanced-configuration), and is definitely why the app won't build locally without a `.env` file present (or setting` NODE_PATH` directly).

For Travis CI, it's set elsewhere [as you can see here](https://travis-ci.org/Coding-Coach/coding-coach/builds/473196327#L435).

DanielRuf:
It is not in https://github.com/Coding-Coach/coding-coach/blob/develop/.travis.yml so still, it could be better. And postinstall should invoke a script which does this.

Try `pnpm install` in the cloned repo to see the issues.

DanielRuf:
Well, did you run the other scripts too?

> Steps to QA
> Run pnpm install && pnpm run lint && pnpm run i18n-extract && pnpm run i18n-compile && pnpm run test && pnpm run build

DanielRuf:
See my initial comment ;-)

DanielRuf:
> I'm seeing the issues you're talking about. `pnpm` doesn't have the flat `node_modules` structure like `yarn` does, so there are some modules that needed to be installed for the commands to work. I see you added them to the `package.json`, so we're basically mimicking `yarn` and adding them back at the top level so that the commands can run. At least, that's my interpretation of it.

Exactly. Another approach is a pnpmfile.js file with a hook.

crysfel:
Thanks for contributing. We are migrating the landing page to gatsby, therefore this codebase is deprecated.

PR #285, opened 2019-01-06
Cookie policy sense-check/grammar-check

ArtOfCode-:
- General grammar check
 - Copy-edited for reading flow
 - Dropped "How we use cookies", because it's just a repeat of "leave it if you're not sure"; the relevant information is in "The Cookies We Set".
 - Removed "GDPR Cookie Policy Generator" text watermark.


trent-boyd:
Very nice proper usage of a semicolon. 👍

PR #291, opened 2019-01-10
Add Dashboard icons

### LINKED ISSUE #259: Components - Icons
image

# Assets
All icons & color codes can be found on the Sketch cloud: https://sketch.cloud/s/xVR5A/paKjK7A

# Details
Given that we are going to use our own custom svg icons, we need to remove `@fortawesome` from our dependencies: https://github.com/Coding-Coach/coding-coach/blob/develop/package.json#L28-L31

We already have an `Icon` [component](https://github.com/Coding-Coach/coding-coach/blob/develop/src/components/icon/Icon.js) to render svg icons, let's use this approach for all new icons.

This task also includes migrating the current icons in the home page from `fortawesome` to the new `Icon` component that we have.

*Edit*: The design contains two states for the icons (active, inactive), that should be handled by the navigation component (We don't have it yet), this task doesn't include that. Only adding the icons to our icons catalog, we might want to create a storybook for all the available icons we already have.

trent-boyd:
# Details

## Description

This adds the dashboard icons, and components for each.

Right now though, I have just added the CC logo and the Dashboard icon. I'd like to get buy-in on this approach before I do the same thing 5–8 times over. 😬

## Steps to QA

1. Run `yarn run storybook`
2. Go to the `Icons` story
3. Let me know what you think?

## Screenshots

image

## Issue

Closes #259


src/components/icon/Dashboard.js
```diff
@@ -0,0 +1,163 @@
+import React from 'react';
+import Icon from './Icon.js';
+
+export default function Dashboard({ active, ...props }) {
+  return (
+    <Icon viewBox="0 0 80 80" {...props}>
+      {active ? (
```
crysfel: seems like the only thing changing is the color. I think we should just set the right color based on the `active` value, this will prevent duplicating the code

src/components/icon/Icons.stories.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import CCLogo from './CCLogo';
+import Dashboard from './Dashboard';
+
+storiesOf('Icons', module)
```
crysfel: Rename to `components/Icons`

src/components/icon/svg/cc-logo.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
crysfel: I don't think we need the SVGs, let's remove them

src/components/icon/Dashboard.js
trent-boyd: I guess that is technically more efficient since the vectors don't change.

src/components/icon/Icons.stories.js
trent-boyd: Will do.

src/components/icon/svg/cc-logo.svg
trent-boyd: I'd like to have them somewhere in source control just in case. Exporting them can be a pain.

crysfel: @emmawedekind added them in the invision board we have

emmawedekind: Hi yes, assets are here: https://projects.invisionapp.com/boards/UF3Q7E6QEZM/

trent-boyd:
@crysfel: Ready for a second look here.

src/components/icon/Dismiss.js
```diff
@@ -0,0 +1,17 @@
...
+        <g transform="translate(19.000000, 19.000000)">
+          <circle fill="#EA7A71" fill-rule="nonzero" cx="21.5" cy="21.5" r="21.5" />
+          <g transform="translate(14.000000, 14.000000)" fill="#FFFFFF">
+            <polygon points="15 1.5 13.5 0 7.5 6 1.5 0 0 1.5 6 7.5 0 13.5 1.5 15 7.5 9 13.5 15 15 13.5 9 7.5" />
+          </g>
+        </g>
+      </g>
+    </Icon>
+  );
+}
```
trent-boyd: On second viewing, we might not even need this since we have a "Close" icon.

nitish24p:
Just a suggestion can you add documentation via a .md file too.. You can see buttons example

trent-boyd:
Hmm. I’m not sure it needs documentation. It has one prop and thorough Storybook examples.

src/components/icon/Dashboard.js
```diff
@@ -0,0 +1,83 @@
...
+
+export default function Dashboard({ active, ...props }) {
+  return (
+    <Icon viewBox="0 0 80 80" {...props}>
+      <g id="dashboard" stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+        <g transform="translate(15.000000, 15.000000)">
+          <path
+            d="M44.1940916,42.8409798 C39.1471856,48.2809744 32.3243291,50.9988316 25.5026399,50.999999 C18.6358126,51.0019444 11.7701527,48.2490682 6.71546422,42.7464286 C-2.32071543,32.9100031 -2.4012636,17.3429259 7.46841436,7.47381612 C12.4526718,2.48984564 18.9774612,-0.00116689225 25.5026399,4.10061094e-07 C32.0247055,0.00116771237 38.5467711,2.49295843 43.5286938,7.47381612 C53.3217149,17.2662734 53.4295015,32.8882136 44.1940916,42.8409798 Z"
+            id="Path"
+            fill={active ? '#69D5B1' : '#BFBFBF'}
```
emmawedekind: I wonder if we could declare `const` variables for the hex codes? @crysfel what do you think?

emmawedekind:
Looks good to me! Let's wait and see what @crysfel thinks! I just left one comment about hex codes :)

src/components/icon/Dashboard.js
crysfel: I think that would be nice to have, that way we can easily update the colors if we decide to update our theme or something. (A search/replace is not a big deal though)

trent-boyd: I think find-replace might be better for now. I can't think of any truly meaningful, reusable names for a variable here. They seem like one-offs to me.

PR #293, opened 2019-01-10
Toggle classNames on body when modal open/close

### LINKED ISSUE #276: When modal is open, user can still scroll the background
If a modal is open, a user should not be able to scroll behind it.

image

pitakill:
# Details

## Description

This PR aims to don't let the scroll of the `body` tag when a modal is opened, and let it back when the modal is closed

## Steps to QA

- [ ] Open/Close modal must hide/show the `body` scrollbar as a visual proof of not let the scroll happen

## Screenshots

screenshot_2019-01-10_11-52-23
screenshot_2019-01-10_11-55-23

## Issue
Closes #276

### DIFF
--- a/src/components/modal/Modal.js
+++ b/src/components/modal/Modal.js
@@ -9,8 +9,15 @@ class Modal extends React.Component {
     document.body.appendChild(this.modal);
   }
 
+  componentDidMount() {
+    document.body.classList.add('overflow-hidden');
+  }
+
   componentWillUnmount() {
-    document.body.removeChild(this.modal);
+    const { body } = document;
+
+    body.removeChild(this.modal);
+    body.classList.remove('overflow-hidden');
   }
 
   render() {


pitakill:
That makes total sense!

Let me change that :-)

src/components/modal/Modal.js
```diff
@@ -6,11 +6,17 @@ class Modal extends React.Component {
   constructor(props) {
     super(props);
     this.modal = document.createElement('div');
-    document.body.appendChild(this.modal);
+    const { body } = document;
+
+    body.classList.add('overflow-hidden');
+    body.appendChild(this.modal);
```
trent-boyd: This may be personal preference, but I generally put anything that's a side effect (i.e. non-React code that doesn't affect component state) inside of a React lifecycle method. `componentDidMount` in this case. Both of these complete instantly though, so I don't think it's a big deal.

crysfel: Agreed, please move this code to: `componentDidMount`

pitakill: I just did that. :-)

I've improved my React (and awareness of my code) skill in this very first PR. And I'm helping to build a wonderful platform.

Thank you guys. You rock!

PR #297, opened 2019-01-15
fix: change primary-dark and primary-darker (#292)

### LINKED ISSUE #292: Insufficient color contrast
# Issue

## Description

The color contrast with the foreground color in the navigation links is not enough in comparison with the background color.

## Screenshot

capture

Can also have a look at this [checker](https://webaim.org/resources/contrastchecker/?fcolor=009D6C&bcolor=69D5B1)

alivenotions:
# Details

## Description

It improves the color contrast for the navigation links on top. Please look at the concerned issue for the discussion.

## Issue

Closes #292


tureey:
It's not directly related to this task but, should `primary-dark` and `primary-darker` be the same? @emmawedekind just to know if we are missing anything or not.

The color is the one you provided btw?

If that's the case, I think everything is OK

alivenotions:
I noticed this too. And the issue is that `primary-dark` is the one that is being used by the navigation buttons. But if I change just that, then `primary-dark` will be darker than `primary-darker`.

tailwind.js
```diff
@@ -53,9 +53,9 @@ let colors = {
   'primary-lightest': '#D5FFF4',
   'primary-lighter': '#E6F5F0',
   'primary-light': '#69D5B1',
-  'primary-dark': '#009d6c',
+  'primary-dark': '#126D4F',
   'primary': '#00bc89',
-  'primary-darker': '#009d6c',
+  'primary-darker': '#126D4F',
```
crysfel: ha.... we have the same value. Please remove one, just make sure to remove all other places where is being used. Thanks!

emmawedekind:
Great @crysfel think this can be merged or still waiting on something?

PR #303, opened 2019-01-28
Update Input Style

### LINKED ISSUE #230: Update Input Style
image

trent-boyd:
# Details

## Description

This adds styles for focused and errored text inputs. In the case of the error styling, an `error<string>` prop will need to be passed into the `TextField` component.

## Steps to QA

1. Run `yarn run storybook`
2. Go to Components > TextInput
3. Check out the stories.

## Issue

Closes #230


src/components/textfield/__tests__/__snapshots__/TextField.test.js.snap
```diff
@@ -2,10 +2,17 @@
 
 exports[`components/textfield/TextField should render the textfield 1`] = `
 <div
-  class="mb-4"
+  class="mb-4 text-secondary-light focus-within:text-primary"
```
trent-boyd: @crysfel: This `focus-within:text-primary` class doesn't work at all. Tailwind doesn't end up generating the CSS class for it. I've followed the instructions on the Tailwind site and added in the `focus-within` variant in `tailwind.js`, but I'm not having any luck. Do you know how to fix it?

tureey:
It looks good🙂 but I would suggest to change the border color from green to red.

I would do to the same with the text but, if you don't want to show a lot of red, we could change it to grey. What do you think?

src/components/textfield/__tests__/__snapshots__/TextField.test.js.snap
tureey: Just to have a little bit of context, have have you tried until now? Have you compiled it? @trent-boyd

trent-boyd: I didn't officially _compile_ it, but in Storybook's dev mode, the `<style>` block that Tailwind generates doesn't have `focus-within` listed anywhere in it. Other pseudo-classes work fine. It's a fairly new feature in Tailwind (2-3 months) and this is the first time we're using it.

trent-boyd:
@tureey: Personally, I think the CC Green color we're using could be easily confused with a "success" state. The focused styling would ideally remain the same regardless of if the text field has an error. I think it's worth considering other options for errored state, especially when you consider red-green color blindness. Maybe some iconography could help.

src/components/textfield/__tests__/__snapshots__/TextField.test.js.snap
crysfel: You need to enable it, for some reason that's not documented O.o Just add `focus-within` to the `textColors` array to `tailwind.js` file

```
textColors: ['responsive', 'hover', 'focus', 'focus-within'],
```

src/components/textfield/TextField.js
```diff
@@ -1,21 +1,39 @@
...
+    `
+      block
+      border rounded-sm
+      leading-tight
+      border-secondary-lighter bg-white
+      w-full p-2
+      focus:outline-none focus:border-primary focus:text-primary
+    `,
+    {
+      ['border-danger']: !!error,
```
crysfel: No need to add the extra brackets:

```
'border-danger': !!error,
```

src/components/textfield/TextField.js
```diff
@@ -1,21 +1,39 @@
...
+    },
+  );
+
   return (
-    <div className="mb-4">
+    <div className="mb-4 text-secondary-light focus-within:text-primary">
       {label && (
-        <label className="block text-secondary-light mb-1" htmlFor={name}>
+        <label className={classnames('block mb-1', { ['text-danger']: !!error })} htmlFor={name}>
```
crysfel: Please remove the extra brackets here too

tailwind.js
```diff
@@ -778,12 +778,12 @@ module.exports = {
   modules: {
     appearance: ['responsive'],
     backgroundAttachment: ['responsive'],
-    backgroundColors: ['responsive', 'hover', 'focus', 'group-hover'],
+    backgroundColors: ['responsive', 'hover', 'focus', 'focus-within', 'group-hover'],
```
crysfel: I think we should remove `focus-within` from the styles we are not using, just to keep the css smaller and with only the things we need.

src/components/textfield/__tests__/__snapshots__/TextField.test.js.snap
trent-boyd: @crysfel: Yep, I've already done that. No dice.

crysfel: That did the trick for me in a different project where I needed to use `focus-within`. Maybe the version we are using doesn't support it? 🤷‍♂️

trent-boyd: @crysfel: I figured it out. As a Tailwind n00b, I didn't know I had to have a process running to generate new styles. `yarn run wach:css` was the step I missed. 😅

PR #306, opened 2019-02-08
As user I want to be able to scroll to the top clicking a button once I am not on the top of the page. Fixes #130

### LINKED ISSUE #130: As user I want to be able to scroll to the top clicking a button once I am not on the top of the page
# Issue
Once the user has scrolled down, he can't go to the top of the page. He has to scroll to the top manually.

Would be nice to have a button that scrolls-top.

## Description

Float button at the bottom-right corner of the screen that scrolls to the top of the page

## Direct Requirements

- [ ] Create the component
- [ ] Functionality to scroll top

kriyeng:
# Details

## Description

Once the user scrolls down the page a button appears in the bottom right of the page. When clicked, the page scrolls using smooth animation to the top of the page.

* Responsive: Based on screen size. When user scrolls more than 25% of the current screen size, the button will appear.
* The component is totally isolated. It can be used in any page. 
* The component can be added at the bottom of any page using `<ScrollTopButton />`. No props needed.

* Created a component named ScrollTopButton. 
* Added the button on the Home.js page
* Modified Modal.js component to add `z-20` to set a `z-index : 20`. Avoids the button to get over the modal. ScrollTopButton has `z-10` class.

## Steps to QA

- [ ] Scroll down the page at more than 25% of the screen height
- [ ] The button has to appear at bottom right. With slow transition and from opacity 0 to 100.
- [ ] Click on the button. The page has to be scrolled to top. The button has to disappear.

- [ ] Scroll the page on small screen
- [ ] Open a large modal.
- [ ] The button has to be behind the modal.

## Screenshots

scrolltopbutton

## Issue

Closes #130


src/pages/home/components/ScrollTopButton.js
```diff
@@ -0,0 +1,59 @@
...
+
+  clickHandler = () => {
+    window.scrollTo({
+      behavior: 'smooth',
+      top: 0,
+    });
+  };
+
+  render() {
+    const styles = {
```
tureey: Since the values of this variable are 'static', what do you think about creating this `const styles` outside the component or as a `static property`?

Doing so, we would avoid the reinitialization every time it re-renders

src/pages/home/components/ScrollTopButton.js
```diff
@@ -0,0 +1,59 @@
...
+        right: -65,
+      },
+      show: {
+        right: 20,
+      },
+    };
+    const { scrollToTopClass } = this.state;
+    return (
+      <div
+        aria-label="Scroll to top of page"
```
tureey: Could we use `i18n`?

kriyeng:
@tureey good points! I've checked and modified both. I hope it is ok now. Thank you!

src/pages/home/components/ScrollTopButton.js
kriyeng: Chenges done.

src/pages/home/components/ScrollTopButton.js
kriyeng: Changes Done!

src/components/modal/Modal.js
```diff
@@ -31,7 +31,7 @@ class Modal extends React.Component {
       <div
         className={classNames(
```
tureey: Since there is no classes depending on variables, I would remove too the function `classNames()`

kriyeng: @tureey one newbie general question. As this is not related to this issue, I should change in this issue? 
Talking in general cases, I added a class to the modal during this issue, but I haven't think/designed how the modal has to be or I don't know if there's another issue related to this problem. I should change here? how it's usually done, I should fix? I should ask before doing it? Of course, I have no problem in doing that! 
Thank you!

tureey: Oh sure! :+1: 

IMHO if changing something that is not related to a task takes some time... We should not do it. There is not point of.. "blocking" a `PR` because of another thing that is not related to what we were doing.

On the other hand. I'm the kind of guy who fixes things every time I open a file (clean up the house):
- Format
- Variable names
- Quick wins
- etc...

In this case, the only thing required is to remove the `classNames()` function and its `import`. Personally I would do it in this `PR` because it will take @crysfel between 4h-12h to approve the `PR`. 

If you can't or even if you want, it's OK since it was not the main issue for this `PR` I'll do it tomorrow or we can open an issue :smile:

kriyeng: Thank you @tureey for your very clear explanation. I understand how to proceed from now on.
If I'm able to detect those mistakes has to be seen! 😆 
I've commited the changes!
Thank you!

tureey: It was not a mistake! :smirk:  well done again!

src/components/modal/Modal.js
```diff
@@ -29,19 +28,12 @@ class Modal extends React.Component {
 
     return (
       <div
-        className={classNames(
-          'fixed pin w-full h-full bg-modal',
-          'flex items-center justify-center',
-        )}
+        className="fixed pin w-full h-full bg-modal flex items-center justify-center z-20"
```
crysfel: Thanks for cleaning this 🙌

src/pages/home/Home.js
```diff
@@ -10,6 +10,7 @@ import LegalModal from './components/LegalModal';
 import Hero from './components/Hero';
 import HomeSection from './components/HomeSection';
 import SocialMedia from './components/SocialMedia';
+import ScrollTopButton from './components/ScrollTopButton';
```
crysfel: Should we move this component to `components`? I think we might want to use it somewhere else and not only on the home page, right?

src/pages/home/components/ScrollTopButton.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import Button from 'components/button/Button';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
crysfel: We are now defining our own icons, check `components/icon`. Please don't use `fortawesome` anymore, we want to remove that dependency. 

Given that we don't have an arrow icon in our library, maybe just use the vectors from `fortawesome` to create our own icon, I didn't see an arrow icon in our set of icons here: https://sketch.cloud/s/xVR5A/ELzbv7q

src/pages/home/Home.js
kriyeng: Sure! I'll move.

src/pages/home/components/ScrollTopButton.js
kriyeng: Good! I'll try to get  the icon from `fontawesome`. Because I have a busy week, I'll got this done for the next weekend! I hope this works for you. Thank you!

src/pages/home/Home.js
kriyeng: Done!

src/pages/home/components/ScrollTopButton.js
kriyeng: Created a new icon in the component/icon folder named ArrowUp based on svg Icon component
On the ScrollTopButton component replaced `fontawesome` with the new ArrowUp button
I hope everything is fine in these changes!

PR #326, opened 2019-06-10
Fix home tests; Add test coverage by default

jpggvilaca:
# Details
Some tests we broken after I just forked and cloned.

## Description
This PR fixes a couple of tests and adds the `coverage` flag by default to the test script

## Notes
Currently the test coverage is ~14% which is not that good. May I suggest adding more tests and snapshots as well?

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build:ssr": "react-snap",
     "deploy": "gh-pages -d build",
     "lint": "eslint src/",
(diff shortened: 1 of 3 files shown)


tureey:
yarn failing @jpggvilaca ?

crysfel:
Thanks for contributing. We are migrating the landing page to gatsby, therefore this codebase is deprecated.

PR #361, opened 2019-11-06
Pagination for blog

alizand1992:
Some notes:
* This is based on master and not develop since blog exist in master.
* I wanted to know if the way I solved this is acceptable before I refactor. please let me know if this is ok so I can start refactoring
* How many posts per page should this show?
* should the page numbers show if there is only one page?
* if the number of pages is more than the width of the screen, this will create a new row. if this is not acceptable, please provide a design for what it should look like if there are many pages.

### DIFF
--- a/src/components/Pagination/index.js
+++ b/src/components/Pagination/index.js
@@ -1,0 +1,49 @@
+import React from 'react';
+
+export default function Pagination({ totalPages, current, setCurrent }) {
+  if (totalPages === 1) {
+    return null;
+  }
+
+  const containerStyle = {
+    display: 'inline-block',
+    marginBottom: '3px',
+    textAlign: 'right',
+    width: '100%',
+  };
+  let pageNumbers = [];
+
+  for (let i = 1; i <= totalPages; i++) {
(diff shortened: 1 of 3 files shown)


alizand1992:
@crysfel Would you be able to look at my commit and notes please? I'm eager to fix any issues that it may have.

crysfel:
I'm so sorry for the late response 😅

answering your questions:

> * This is based on master and not develop since blog exist in master.
Yes, we should delete develop. Will do that soon

> * How many posts per page should this show?
Maybe 10 would be a good number

> * should the page numbers show if there is only one page?
No, just hide the component

> * if the number of pages is more than the width of the screen, this will create a new row. if this is not acceptable, please provide a design for what it should look like if there are many pages.
For now is fine... we can deal with that when the time comes 😬

alizand1992:
@crysfel Awesome. I will make the changes

crysfel:
@alizand1992 ping me on slack if something is not clear or if you need help

OmarMoataz:
hey @crysfel, is this feature still under development? Can I pick up Alizand left off?

crysfel:
@OmarMoataz  That would be awesome!

OmarMoataz:
I can start working on this the week after next just so you know. :)

PR #371, opened 2020-03-11
New mock openning

shayanypn:


crysfel:
Please checkout from `master`, develop is very different and we should delete it

shayanypn:
I created another PR for this change.

PR #434, opened 2023-06-21
Fix: aria-labels for static and dynamic links plus aria-labelledby, titles and descriptions for svg images

ttkapostol:
Hello!

We have tried to improve the accessibility of the page by adding aria-labels for static and dynamic links plus aria-labelledby, titles and descriptions for svg images.

We hope to here from you soon!


ttkapostol:
PD. We know that the checks are failing due to some plugins and dependencies we don´t have. We hope that is not a problem.

